PR #14, opened 2018-06-06
Update a few more things

Byron:
I am about to replace my own little update-shellscript with topgraph as it offers better reporting and overall, is the higher-quality way of doing it.

Probably I will just keep pushing into this PR until it's done, or you tell me to stop :D.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -268,6 +268,16 @@ fn main() -> Result<(), Error> {
             .report("Node Package Manager", &mut reports);
     }
 
+    if let Ok(apm) = which("apm") {
+        terminal.print_separator("Atom Package Manager");
+        Command::new(&apm)
+            .args(&["upgrade", "--confirm=false"])
+            .spawn()?
+            .wait()
+            .map_err(Error::from)?
+            .report("Atom Package Manager", &mut reports);
+    }
+
     let mut reports: Vec<_> = reports.into_iter().collect();
     reports.sort();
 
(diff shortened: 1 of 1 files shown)


r-darwish:
Thanks for the pull request :)

I was actually wondering about whether to add support for rustup. There are two things to consider.

* The stable channel - It's safe to update. However, Rust versions don't release often, and when they do, people are usually aware of that, so maybe it's a bit of a waste to try to update something which doesn't update very often?

* The nightly channel - This actually updates every night, but has the potential for breaking things. I don't use nightly personally so I'm not sure what's the desired workflow for people who actually use nightly. Do such people update nightly every day, or do they only do that whenever they feel ready?

r-darwish:
Also, could you please remove the "Update from more thing" from the commit messages?

Byron:
Thanks for the hints! I agree, rustup isn't that clear, and personally I usually ended up running it myself for stable. Nightly is a different beast entirely - so the commit is removed.
Also I have updated the subject lines accordingly.
Thanks for making this tool - I slowly get the hang of writing Rust instead of bash to have proper quality for important tooling.

r-darwish:
Awesome, thanks!

I opened up a discussion for Rustup [here](https://github.com/r-darwish/topgrade/issues/16)

r-darwish:
Note that I moved NPM and APM before the system upgrade. I want the system upgrade to be last.

In addition, I removed the first call to NPM as it a local call and not a global one.

PR #45, opened 2018-07-03
Add neovim

tversteeg:
I am using the "app_dirs" crate to discover `~/.config` on Unix and `%APPDATA%/` on Windows.


src/vim.rs
```diff
@@ -50,3 +51,25 @@ pub fn vimrc() -> Option<PathBuf> {
...
+        let mut nvimrc = get_data_root(AppDataType::UserConfig).unwrap();
+        nvimrc.push("nvim/init.vim");
+
+        if nvimrc.exists() {
+            return Some(nvimrc);
+        }
+    }
+
+    {
+        let mut nvimrc = get_data_root(AppDataType::UserCache).unwrap();
```
r-darwish: Is it really possible that the configuration be placed here?

r-darwish:
I don't the the Github code review interface very well so I keep deleting my comment by mistake, so I'll just write it here. Do we really need `run_nvim`? Can't we use `run_vim` for Neovim as well?

tversteeg:
Somehow the separate `-e -s` flags instead of `-es` don't work for me in neovim.

r-darwish:
I tried using the current `run_vim` with Neovim 0.3.0 and a configuration based on Plug, which seems to work. Which problem did you encounter?

tversteeg:
Oh now it has the same behavior, maybe it was something else..

PR #62, opened 2018-09-03
Add OCaml package manager (OPAM)

claui:
This adds support for the [OCaml package manager](https://opam.ocaml.org/) (OPAM).

The command line `opam update` first performs a self-update if necessary. Then it updates its package list.
Finally, `opam upgrade` upgrades all locally installed packages.


claui:
> Can you please update the README as well?

Done. Thanks for the review!

PR #75, opened 2018-10-09
Ctrl+C handling

r-darwish:
As stated [here](https://doc.rust-lang.org/std/io/trait.BufRead.html#errors-1), `read_until` (and `read_line`) ignore Ctrl+C, so Topgrade does not respond to Ctrl+C in the retry prompt, and instead will exit only when enter is pressed after Ctrl+C. This is undesirable, so this pull request is a WIP until we find a solution.


src/main.rs
```diff
@@ -56,24 +64,47 @@ struct StepFailed;
...
-        if !terminal.should_retry() {
-            return Some((key, success));
+        let running = ctrlc::running();
+        if !running {
+            ctrlc::set_running(true);
+        }
+
+        let should_retry = execution_context.terminal.should_retry(running);
+
+        if !ctrlc::running() {
```
drrlvn: Maybe `should_retry()` should return `Err` if it was interrupted instead of returning `false`.

src/ctrlc.rs
```diff
@@ -0,0 +1,29 @@
...
+}
+
+pub fn set_handler() {
+    let sig_action = signal::SigAction::new(
+        signal::SigHandler::Handler(handle_sigint),
+        signal::SaFlags::empty(),
+        signal::SigSet::empty(),
+    );
+    unsafe {
+        signal::sigaction(signal::SIGINT, &sig_action).unwrap();
```
drrlvn: The `unwrap()` call can be outside the `unsafe` block.

src/terminal.rs
```diff
@@ -84,14 +86,20 @@ impl Terminal {
             let _ = self.stdout.reset();
             let _ = self.stdout.flush();
 
-            if let Ok(c) = Term::stdout().read_char() {
+            let answer = Term::stdout().read_char();
```
drrlvn: Use `?` instead of manually checking return value.

src/terminal.rs
```diff
@@ -84,14 +86,20 @@ impl Terminal {
...
+            if !ctrlc::running() {
+                return Err(Interrupted);
+            }
+
+            if let Ok(c) = answer {
                 match c {
-                    'y' | 'Y' => return true,
-                    'n' | 'N' | '\r' | '\n' => return false,
+                    'y' | 'Y' => return Ok(true),
+                    'n' | 'N' | '\r' | '\n' => return Ok(false),
```
drrlvn: Remove `'\r'` as it cannot be returned by `read_char()`.

PR #90, opened 2018-11-17
Adding a new flag for cleaning up old resources

nhellwig:
If you are using applications like brew they may keep a history of
packages downloaded. Especially the bigger your number of installed
packages for homebrew gets the bigger those directories get. This can
quickly add up to a couple of GBs.

In my current workflow of updating, I was always cleaning up homebrews / linuxbrews "unneeded" resources after running it.

This pull-request adds a new flag to topgrade that enables users to explicitly ask for removing those kind of files. The PR contains an implementation for "brew" but it can easily spread to other commands as well.


r-darwish:
Awesome! Thank you.

PR #98, opened 2018-12-09
Add `--no-vim` flag

woohgit:
Just like we have the `--no-emacs` flag, it would be nice to have the `--no-vim` flag, to avoid the upgrade of the vim/neovim configuration files and packages.


r-darwish:
Thanks!

Issue #105, opened 2019-01-05
Generic blacklist g


r-darwish:

nCrazed:
Is this intended to be a replacement for the specific `--no-system`, `--not-git-repos`, `--no-vim`, `--no-emacs` flags, or something else entirely?

r-darwish:
Yeah, I think it makes more sense that we'll have a `--disable emacs,vim` instead of `--no-vim --no-emacs`, etc...

PR #108, opened 2019-01-13
WIP: Generic Blacklist implementation

nCrazed:
A proof of concept implementation for #105 

I am using `enum_derive` here just focus on the actual important changes. The final implementation should separate enum type names from the CLI options to allow use of conventional enum naming and possibly many to one mapping between CLI options and enums in the future.


r-darwish:
Thanks for doing that. `custom_derive` doesn't seem to be maintained anymore. Is there any more maintained alternative?

nCrazed:
I don't think we need any alternatives, just implement the necessary string to enum mapping logic explicitly.

I just didn't want to go trough the effort before confirming that I understood the issue correctly.

nCrazed:
@r-darwish thoughts on using [lazy-static](https://github.com/rust-lang-nursery/lazy-static.rs) to statically initialize a map of String -> Group mappings that can be used for both the CLI arg mapping to enums as well as automatically generating the full available flag list in the error message?

src/config.rs
```diff
@@ -10,6 +10,39 @@ use toml;
 
 type Commands = BTreeMap<String, String>;
 
+#[derive(Debug, PartialEq)]
```
r-darwish: I think this can derive `Copy` and it will prevent the need to pass it as reference in `main.rs`

src/config.rs
```diff
@@ -10,6 +10,39 @@ use toml;
...
+pub enum Group {
```
r-darwish: I rather that we call it `Steps`

r-darwish:
```patch
@@ -1,16 +1,25 @@
 use super::error::{Error, ErrorKind};
 use directories::BaseDirs;
 use failure::ResultExt;
+use lazy_static::lazy_static;
 use serde::Deserialize;
 use shellexpand;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 use std::fs;
 use structopt::StructOpt;
 use toml;
 
 type Commands = BTreeMap<String, String>;
 
-#[derive(Debug, PartialEq)]
+lazy_static! {
+    static ref STEPS_MAPPING: HashMap<&'static str, Group> =
+        [("system", Group::System), ("git-repos", Group::GitRepos)]
+            .iter()
+            .cloned()
+            .collect();
+}
+
+#[derive(Debug, PartialEq, Copy, Clone)]
 pub enum Group {
     /// Don't perform system upgrade
     System,
@@ -24,22 +33,17 @@ pub enum Group {
     Gem,
 }
 
+impl Group {
+    fn possible_values() -> Vec<&'static str> {
+        STEPS_MAPPING.keys().cloned().collect()
+    }
+}
+
 impl std::str::FromStr for Group {
     type Err = structopt::clap::Error;
+
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        Ok(match s {
-            "system" => Group::System,
-            "git-repos" => Group::GitRepos,
-            "vim" => Group::Vim,
-            "emacs" => Group::Emacs,
-            "gem" => Group::Gem,
-            _ => {
-                return Err(structopt::clap::Error::with_description(
-                    "Allowed values: system, git-repos, vim, emacs, gem",
-                    structopt::clap::ErrorKind::InvalidValue,
-                ));
-            }
-        })
+        Ok(STEPS_MAPPING.get(s).unwrap().clone())
     }
 }
 
@@ -101,7 +105,7 @@ pub struct Opt {
     #[structopt(long = "no-retry")]
     pub no_retry: bool,
 
-    /// Do not perform upgrades for the given groups. Allowed options: system, git-repos, vim, emacs
-    #[structopt(long = "disable")]
+    /// Do not perform upgrades for the given groups
+    #[structopt(long = "disable", raw(possible_values = "&Group::possible_values()"))]
     pub disable: Vec<Group>,
 }
```

@nCrazed this seems to do what we want

src/config.rs
nCrazed: Doesn't look like that is the case: https://doc.rust-lang.org/src/core/slice/mod.rs.html#1283-1287

I've tried various combinations of `Copy` and `Clone` with no luck.

src/config.rs
nCrazed: Yeah, that's probably a better name

nCrazed:
Oh, I missed the fact that `lazy_static` is already in the dependencies :D

src/config.rs
nCrazed: That being said, we still need it to be at least `Clone` for the new `FromStr` implementation.

PR #116, opened 2019-01-30
Normalize paths in Windows (fix #113)

### LINKED ISSUE #113: Paths are displayed weirdly on Windows
There's a `//?/` before the path, e.g.:

```
―― Pulling //?/C:/Users/Dror Levin/config-files ―――――――――――――――――――――――――――――――――
```

Also `topgrade` is printing slashes instead of backslashes but that's acceptable.

r-darwish:
While `//?/C:/Users/Someone/config-files` is a valid path, it's not very human friendly.
Such paths are returned from `BaseDirs`. This patch formats them in the traditional way.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,8 +1,8 @@
 use super::error::{Error, ErrorKind};
 use log::{debug, error};
 use std::ffi::OsStr;
-use std::fmt::Debug;
-use std::path::{Path, PathBuf};
+use std::fmt::{self, Debug};
+use std::path::{Component, Path, PathBuf};
 use std::process::{ExitStatus, Output};
 use which_crate;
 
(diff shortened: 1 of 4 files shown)


src/utils.rs
```diff
@@ -68,3 +68,44 @@ pub fn which<T: AsRef<OsStr> + Debug>(binary_name: T) -> Option<PathBuf> {
...
+            let mut iterator = self.path.components().peekable();
+
+            while let Some(component) = iterator.next() {
+                match &component {
+                    Component::Normal(c) if *c == "?" => (),
+                    Component::Normal(c) => {
+                        write!(f, "{}", c.to_string_lossy())?;
+
+                        if iterator.peek().is_some() {
+                            write!(f, "\\")?;
```
drrlvn: Use `std::path::MAIN_SEPARATOR`.

src/utils.rs
```diff
@@ -68,3 +68,44 @@ pub fn which<T: AsRef<OsStr> + Debug>(binary_name: T) -> Option<PathBuf> {
...
+                        }
+                    }
+                    Component::RootDir => (),
+                    _ => unimplemented!("{:?}", component),
```
drrlvn: Handle `Component::Prefix`.

src/utils.rs
```diff
@@ -68,3 +68,58 @@ pub fn which<T: AsRef<OsStr> + Debug>(binary_name: T) -> Option<PathBuf> {
...
+    }
+}
+
+impl<'a> fmt::Display for HumanizedPath<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if cfg!(windows) {
+            let mut iterator = self.path.components().peekable();
+
+            while let Some(component) = iterator.next() {
+                let is_prefix = if let Component::RootDir = &component {
```
drrlvn: Why is this needed? How can both `is_prefix` and `printed` be `true`?

src/utils.rs
```diff
@@ -68,3 +68,88 @@ pub fn which<T: AsRef<OsStr> + Debug>(binary_name: T) -> Option<PathBuf> {
         }
     }
 }
+
+/// `std::fmt::Display` implementation for `std::path::Path`.
+///
+/// This struct differs from `std::path::Display` in that in Windows it takes care of printing slashes
```
drrlvn: This is backwards - on Windows it prints backslashes instead of slashes.

PR #130, opened 2019-02-26
fix(zplug): search appropriate paths for zplug folder

ericmarkmartin:
Check ZPLUG_HOME environment variable in addition to ~/.zplug. Also consider ZDOTDIR when sourcing
.zshrc.

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -22,7 +22,7 @@ fn get_zshrc(base_dirs: &BaseDirs) -> Result<String, ()> {
         Err(_) => Err(()),
     };
     zshrc
-        .unwrap_or(base_dirs.home_dir().join(".zshrc"))
+        .unwrap_or_else(|_| base_dirs.home_dir().join(".zshrc"))
         .to_str()
         .map(|s| s.to_owned())
         .ok_or(())
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -8,20 +8,15 @@ use std::path::Path;
 use std::process::Command;
 
 fn zplug_exists(base_dirs: &BaseDirs) -> bool {
-    let home_exists = match env::var("ZPLUG_HOME") {
-        Ok(ref zplug_home) => Path::new(zplug_home).exists(),
-        Err(_) => false,
-    };
-    let dotdir_exists = base_dirs.home_dir().join(".zplug").exists();
-    home_exists || dotdir_exists
+    env::var("ZPLUG_HOME")
+        .map(|ref zplug_home| Path::new(zplug_home).exists())
+        .map(|p| p || base_dirs.home_dir().join("zplug").exists())
+        .unwrap_or(false)
 }
 
 fn get_zshrc(base_dirs: &BaseDirs) -> Result<String, ()> {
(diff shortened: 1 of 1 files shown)


src/steps/os/unix.rs
```diff
@@ -1,21 +1,42 @@
...
+    let home_exists = match env::var("ZPLUG_HOME") {
+        Ok(ref zplug_home) => Path::new(zplug_home).exists(),
+        Err(_) => false,
+    };
+    let dotdir_exists = base_dirs.home_dir().join(".zplug").exists();
+    home_exists || dotdir_exists
+}
+
+fn get_zshrc(base_dirs: &BaseDirs) -> Result<String, ()> {
+    let zshrc = match env::var("ZDOTDIR") {
```
r-darwish: You can use [ok_or_else](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else)

src/steps/os/unix.rs
```diff
@@ -1,21 +1,42 @@
...
 use crate::executor::{CommandExt, RunType};
 use crate::terminal::print_separator;
 use crate::utils::{require, which};
 use directories::BaseDirs;
+use std::env;
 use std::path::Path;
 use std::process::Command;
 
+fn zplug_exists(base_dirs: &BaseDirs) -> bool {
+    let home_exists = match env::var("ZPLUG_HOME") {
```
r-darwish: You can use some combination of option methods to avoid matching

ericmarkmartin: You're absolutely right—It's late in my timezone. Just pushed changes.

ericmarkmartin:
Just pushed the requested changes. Also just wanted to thank you for this tool—it really makes my life a lot easier

src/steps/os/unix.rs
```diff
@@ -1,21 +1,37 @@
...
 use crate::executor::{CommandExt, RunType};
 use crate::terminal::print_separator;
 use crate::utils::{require, which};
 use directories::BaseDirs;
+use std::env;
 use std::path::Path;
 use std::process::Command;
 
+fn zplug_exists(base_dirs: &BaseDirs) -> bool {
+    env::var("ZPLUG_HOME")
```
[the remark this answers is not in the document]
ericmarkmartin: You're probably right on that—I was just writing some Haskell and thinking in terms of monoids. I'll make those changes but how would you feel about my writing up some utility module that makes working with a bunch of paths easier? Specifically, I'm thinking it would allow you to chain a bunch of strings and then take the first one that's a valid path or something.

src/steps/os/unix.rs
```diff
@@ -1,21 +1,37 @@
...
+        .map(|ref zplug_home| Path::new(zplug_home).exists())
+        .map(|p| p || base_dirs.home_dir().join("zplug").exists())
+        .unwrap_or(false)
+}
+
+fn get_zshrc(base_dirs: &BaseDirs) -> Result<String, ()> {
```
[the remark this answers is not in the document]
ericmarkmartin: I expect by `Result<PathBuf>` you mean `Result<PathBuf, ()>`?

PR #134, opened 2019-03-10
feat(zplug): print separator for zplug step

ericmarkmartin:


src/steps/os/unix.rs
```diff
@@ -9,6 +9,7 @@ use std::process::Command;
 
 pub fn run_zplug(base_dirs: &BaseDirs, run_type: RunType) -> Result<(), Error> {
     let zsh = require("zsh")?;
+    print_separator("zplug");
```
r-darwish: Oops. I probably forgot that.

Anyway, separators should be printed after all requirements have been made. The current implementation will print the zplug separator even if you only have zsh install and don't use zplug. You should move that line before the cmd line.

r-darwish: You still need to remove this one

ericmarkmartin:
Sorry about that. That's what you get for late night coding, I suppose. Should be fixed now.

PR #147, opened 2019-05-14
Fix arch based distribution identification

idanski:


src/steps/os/linux.rs
```diff
@@ -41,6 +41,7 @@ impl Distribution {
                 (_, Some("debian")) | (_, Some("ubuntu")) => Distribution::Debian,
                 (_, Some("\"suse\"")) => Distribution::Suse,
                 ("arch", _) => Distribution::Arch,
+                (_, Some("archlinux")) => Distribution::Arch,
```
r-darwish: Please join it with the previous line with the `|` operator

PR #149, opened 2019-05-15
Adds support for 'myrepos' cli tool

derphilipp:


src/steps/generic.rs
```diff
@@ -88,6 +88,27 @@ pub fn run_pipx_update(run_type: RunType) -> Result<(), Error> {
     run_type.execute(&pipx).arg("upgrade-all").check_run()
 }
 
+pub fn run_myrepos_update(base_dirs: &BaseDirs, run_type: RunType) -> Result<(), Error> {
+    let myrepos = utils::require("mr")?;
+    base_dirs.home_dir().join(".mrconfig").require()?;
+
+    print_separator("myrepos");
+
+    match base_dirs.home_dir().to_str() {
```
r-darwish: This made me think about `to_str` for paths. We shouldn't do that at all, but instead use the paths as is with `.arg()`. I made a [patch](https://github.com/r-darwish/topgrade/commit/93d3eb2) that fixes the other places. Place change this one as well.

derphilipp: Good point, this is way better!

derphilipp:
Thank you! 👍

PR #154, opened 2019-05-16
Adds os recognition support for manjaro linux

derphilipp:


src/steps/os/linux.rs
```diff
@@ -36,7 +36,7 @@ impl Distribution {
             match (os_release.id.as_ref(), os_release.id_like.as_ref().map(String::as_str)) {
                 ("debian", _) | (_, Some("debian")) | (_, Some("ubuntu")) => Distribution::Debian,
                 (_, Some("\"suse\"")) => Distribution::Suse,
-                ("arch", _) | (_, Some("archlinux")) | (_, Some("\"arch\"")) => Distribution::Arch,
+                ("arch", _) | (_, Some("archlinux")) | (_, Some("\"arch\"")) | ("manjaro", _) => Distribution::Arch,
```
idanski: I think this line is not needed, Manjaro should be covered the the ID_LIKE "arch", no?

r-darwish: The real problem is that serde-ini treats `ID_LIKE="arch"` and `ID_LIKE=arch` as different result as it does not strip quotes. I'll open a bug for them.

@derphilipp - In the meantime I prefer that `(_, Some("arch")) | (_, Some("\"arch\""))` instead of checking `ID`

derphilipp: Ah, thank for the notice, I have fixed that according to @r-darwish s suggestion

PR #166, opened 2019-06-12
added support for SDKMAN! updates and upgrades

eltonlika:
- Implemented SDKMAN self-update
- Implemented SDKMAN candidates update
- Implemented SDKMAN packages upgrade


eltonlika:
not passing cargo fmt check; will request again after applying fixes

PR #167, opened 2019-06-12
Implemented support for SDKMAN! updates and upgrades

eltonlika:
- Implemented SDKMAN self-update
- Implemented SDKMAN candidates update
- Implemented SDKMAN packages upgrade


src/config.rs
```diff
@@ -43,6 +44,8 @@ pub enum Step {
     Emacs,
     /// Don't upgrade ruby gems
     Gem,
+    /// Don't upgrade SDKMAN! and it's packages
```
r-darwish: "its packages"

src/main.rs
```diff
@@ -468,6 +468,21 @@ fn run() -> Result<(), Error> {
         }
     }
 
+    if config.should_run(Step::Sdkman) {
```
r-darwish: As I understand, SDKMAN! doesn't work on Windows at all, so this whole if should be inside `#[cfg(unix)]`

src/steps/generic.rs
```diff
@@ -151,3 +152,43 @@ pub fn run_remote_topgrade(run_type: RunType, hostname: &str) -> Result<(), Erro
...
+        .map(PathBuf::from)
+        .unwrap_or_else(|_| base_dirs.home_dir().join(".sdkman"))
+        .join("bin")
+        .join("sdkman-init.sh")
+        .require()
+        .map(|p| format!("{}", &p.display()))?;
+
+    print_separator("SDKMAN!");
+
+    let cmd_update = format!(
```
r-darwish: Do we really have to execute the shell ourselves? I use fish as my shell so `source` will probably fail on my computer. If there's no way other then to execute the shell ourselves then I suggest that we always execute bash, as that's the shell SDKMAN is written in.

src/main.rs
eltonlika: Yes, initially I thought so but I think that there are ways to make it run under cygwin and that's why I removed the conditional compilation flag.

src/steps/generic.rs
r-darwish: I think that if Topgrade is compiled in Cygwin then Rust will set the `unix` configuration instead of `windows`. That's the assumption that I wrote the other steps with.

src/main.rs
r-darwish: I think that if Topgrade is compiled in Cygwin then Rust will set the `unix` configuration instead of `windows`. That's the assumption that I wrote the other steps with.

src/steps/generic.rs
eltonlika: Well, SDKMAN has no binary, but is invoked by a bash function sourced from sdkman-init.sh, so I'm out of ideas when it comes to make it run without executing shell.
Any suggestions?
Otherwise I will change the code to check for bash and execute every command in it's statement.

src/main.rs
eltonlika: But what if topgrade is compiled on the windows toolchain and then invoked from inside cygwin? Should it not be expected to work?

src/steps/generic.rs
r-darwish: I see. If that's the case then I suggest that we always use bash instead of `SHELL`, because `SHELL` might point to a shell that's not bash compatible, such as fish.

src/main.rs
r-darwish: I don't think so. If you install a native Python and launch it from cygwin, you won't be able to do `import signal` from that Python, because it's still a native Python. If someone wishes to use Topgrade inside Cygwin then they should compile it with the Rust version that comes with Cygwin, and in that case I believe the toolchain will be considered as unix.

src/steps/generic.rs
```diff
@@ -151,3 +152,24 @@ pub fn run_remote_topgrade(run_type: RunType, hostname: &str) -> Result<(), Erro
...
+        .map(PathBuf::from)
+        .unwrap_or_else(|_| base_dirs.home_dir().join(".sdkman"))
+        .join("bin")
+        .join("sdkman-init.sh")
+        .require()?;
+
+    let shell = env::var("SHELL").unwrap_or_else(|_| String::from("sh"));
+
+    let cmd = format!(
+        "source {} && sdk selfupdate && sdk update && sdk upgrade",
```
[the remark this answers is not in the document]
eltonlika: I changed the calls to be on their own statement.

src/steps/generic.rs
```diff
@@ -151,3 +152,24 @@ pub fn run_remote_topgrade(run_type: RunType, hostname: &str) -> Result<(), Erro
         ])
         .check_run()
 }
+
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
eltonlika: I put the run_sdkman function in unix.rs

src/main.rs
eltonlika: Added the conditional compilation flag to the run_sdkman call.

src/steps/os/unix.rs
```diff
@@ -84,3 +84,51 @@ pub fn run_pearl(run_type: RunType) -> Result<(), Error> {
 
     run_type.execute(&pearl).arg("update").check_run()
 }
+
+pub fn run_sdkman(base_dirs: &BaseDirs, cleanup: bool, run_type: RunType) -> Result<(), Error> {
+    let bash_path = require("bash").map(|p| format!("{}", &p.display()))?;
```
r-darwish: I think you can remove the `.map` and have it stay as `PathBuf`

eltonlika: Yes you are right. I'll remove it and make a new pull request.

PR #176, opened 2019-06-26
Add DragonFly BSD support

strangelittlemonkey:
DragonFly BSD supports the same pkg upgrade and audit of FreeBSD, though the OS upgrade process is different.


r-darwish:
Always happy to see BSD users interested in this project 😄

PR #183, opened 2019-07-31
Show what's new in Git repositories

r-darwish:
image

### DIFF
--- a/src/steps/git.rs
+++ b/src/steps/git.rs
@@ -23,18 +23,22 @@ pub struct Repositories<'a> {
     repositories: HashSet<String>,
 }
 
-fn get_head_revision(git: &Path, repo: &str) -> Option<String> {
+/// Checks whether the latest pull command actually pulled something
+fn did_pull_change(git: &Path, repo: &str) -> bool {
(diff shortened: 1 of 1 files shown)


src/steps/git.rs
```diff
@@ -83,26 +97,53 @@ impl Git {
             .map(|repo| {
                 let repo = repo.clone();
                 let path = format!("{}", HumanizedPath::from(std::path::Path::new(&repo)));
-                println!("{} {}", style("Pulling").cyan(), path);
+                let before_revision = get_head_revision(git, &repo);
```
drrlvn: Instead of manually getting the revision before and after you can use `git log ORIG_HEAD..`, see https://stackoverflow.com/a/6535251.

r-darwish: I also use the if statement to determine whether I should write "Up-to-date" or "Changed":

https://github.com/r-darwish/topgrade/blob/064dab59e972dc534301d44ee3fb6d276bdf88ef/src/steps/git.rs#L118

How could I do that with ORIG_HEAD?

PR #186, opened 2019-08-13
Shellcheck ci scripts, add a small .editorconfig

filalex77:


r-darwish:
Thank you. These scripts were taken from https://github.com/japaric/trust. Perhaps you could also open a pull request there.

filalex77:
@r-darwish Thanks for pointing that out.

japaric/trust#122

PR #197, opened 2019-08-16
Update development packages using yay

severinkaderli:
This change also updates development packages when using yay.


r-darwish:
Thanks!

PR #202, opened 2019-08-20
os: add support for Exherbo

Keruspe:


src/steps/os/linux.rs
```diff
@@ -77,17 +79,20 @@ impl Distribution {
...
             Distribution::Void => upgrade_void(&sudo, run_type),
             Distribution::Solus => upgrade_solus(&sudo, run_type),
+            Distribution::Exherbo => upgrade_exherbo(&sudo, cleanup, run_type),
         }
     }
 
-    pub fn show_summary(self) {
-        if let Distribution::Arch = self {
-            show_pacnew();
+    pub fn show_summary(self) -> Result<(), Error> {
```
r-darwish: I prefer that this function will not fail-able. You can still have `show_eclectic` return `Result` but instead of propagating it in `show_summary` just print it instead.

Keruspe: I simplified this part

r-darwish:
Thanks!

PR #209, opened 2019-08-27
Add a remote_timeout option

fredmorcos:
The default connection timeout for ssh is quite high, with `remote_timeout` the user can choose a shorter timeout in seconds.


r-darwish:
Thank you for the contribution. I think it's best that we'll make a flag called `ssh_arguments` instead in order to give the user more flexibility.

fredmorcos:
> I think it's best that we'll make a flag called ssh_arguments instead in order to give the user more flexibility.

Should I work on this?

r-darwish:
That would be great

PR #211, opened 2019-08-29
Add an ssh_arguments configuration file field

fredmorcos:
Can be used to add SSH command-line arguments like `-o
ConnectTimeout=2`


r-darwish:
Thanks

PR #212, opened 2019-08-30
Fix typo for keep option

jrgarcia:
The current documentation for this option does not make sense and the
documentation for the keep_at_end function differs. I made this change
assuming is a typo.


r-darwish:
Thanks

PR #216, opened 2019-09-04
Add zr support

jedahan:
[zr](https://github.com/jedahan/zr) is a simple zsh package manager, written in rust.

I tried to keep this as similar as possible to the other update mechanism, but do have a few questions for the future:

1. Should we use deeper integration by bringing making `zr` a library and bringing it in as a dependency in the future?
2. Should we really be sourcing the entirety of zshrc when doing shell stuff? Feels side-effecty, but if thats the best way to have the right paths for everything, *shrug*?
3. Should I just make the above two questions issues to discuss separately from this PR?

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ Just run `topgrade`. It will run the following steps:
   * ~/.config/i3
   * Powershell Profile
   * Custom defined paths
+* **Unix**: Run [zr](https://github.com/jedahan/zr) update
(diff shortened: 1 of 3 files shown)


r-darwish:
Thank you and good luck with zr. I'm happy to see a package manager which isn't written in shell scripting. I don't think that there's an advantage linking zr with Topgrade, so I prefer executing it just like any other package manager. As for sourcing the shell, this is what I do with zplug.

PR #221, opened 2019-09-07
Fix grammar

tuxiqae:
"No sudo or yay" => "Not sudo nor yay"


tuxiqae:
Oops, my bad :speak_no_evil:

PR #232, opened 2019-09-29
Added support for Voom for Vim packages

hugoh:
https://github.com/airblade/voom

### DIFF
--- a/src/steps/vim.rs
+++ b/src/steps/vim.rs
@@ -117,3 +117,11 @@ pub fn upgrade_neovim(base_dirs: &BaseDirs, run_type: RunType) -> Result<(), Err
     print_separator(&format!("Neovim ({:?})", plugin_framework));
     upgrade(&nvim, &nvimrc, plugin_framework, run_type)
 }
+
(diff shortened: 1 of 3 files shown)


src/steps/vim.rs
```diff
@@ -117,3 +117,11 @@ pub fn upgrade_neovim(base_dirs: &BaseDirs, run_type: RunType) -> Result<(), Err
     print_separator(&format!("Neovim ({:?})", plugin_framework));
     upgrade(&nvim, &nvimrc, plugin_framework, run_type)
 }
+
+pub fn run_voom(_base_dirs: &BaseDirs, run_type: RunType) -> Result<(), Error> {
+    let zsh = require("voom")?;
```
r-darwish: Please change the variable name to `vim`

src/steps/vim.rs
```diff
@@ -117,3 +117,11 @@ pub fn upgrade_neovim(base_dirs: &BaseDirs, run_type: RunType) -> Result<(), Err
...
+
+    print_separator("voom");
+
+    run_type.execute(zsh).args(&["update"]).check_run()
```
r-darwish: Use `arg` instead of `args`

PR #233, opened 2019-10-02
Support yay arguments in config file (Closes #220)

### LINKED ISSUE #220: Ability to disable yay's --devel flag
Hi @r-darwish,
I would love it if it'd be possible to disable yay's --devel flag using both the CLI and the config file.

P.S. why would I even want to use --devel on daily basis as a regular user?

idanski:
@r-darwish Should I add tests somewhere? didn't figure out where.. :)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -61,6 +61,7 @@ pub struct ConfigFile {
     git_arguments: Option<String>,
     set_title: Option<bool>,
     assume_yes: Option<bool>,
+    yay_arguments: Option<String>,
 }
 
 impl ConfigFile {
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -286,4 +287,13 @@ impl Config {
     pub fn yes(&self) -> bool {
         self.config_file.assume_yes.unwrap_or(self.opt.yes)
     }
+
+    /// Extra yay arguments
+    pub fn yay_arguments(&self) -> String {
```
r-darwish: Can we return `&str`, or will cause a lifetime hell?

src/steps/os/linux.rs
```diff
@@ -135,7 +145,7 @@ fn upgrade_arch_linux(sudo: &Option<PathBuf>, cleanup: bool, run_type: RunType,
             .arg("--pacman")
             .arg(pacman)
             .arg("-Syu")
-            .arg("--devel")
+            .arg(yay_arguments)
```
r-darwish: I don't think this will work when multiple arguments are given. We'll have to split it by space and pass it to `.args`. See [this example](https://github.com/r-darwish/topgrade/blob/ab318905a116d26e1c0ac6c86ef496cb94c8a8f3/src/steps/git.rs#L116)

src/config.rs
r-darwish: put `#[cfg(target_os = "linux")]` so that we won't get an unused warning on

r-darwish:
The only place that Topgrade has tests is the Linux distribution detection, so no tests are needed.

r-darwish:
Thanks!

PR #237, opened 2019-10-09
Allow most command-line options to be configured in config file (closes #188)

### LINKED ISSUE #188: Allow all command-line options to be configured in config file
First of all, thank you for your work on topgrade! It's an incredibly useful tool.

I have one request: it would be nice to allow configuring  all command-line options from the config file. As far as I can tell, at the moment `disable` is the only one that can be specified both in the config file or in  the command line. It would be useful  to allow all others as well, this would make it easier to configure the default behavior of  the tool (e.g. with `tmux`, `no-retry`, `keep`, etc.)

Thanks!

idanski:
@r-darwish There are some parts I'm not 100% sure about the correct behavior (mostly allowed steps), let me know what you think, we can change some implementation details.

But it works :)

Notes: 
* We lose ConfigFile::read(base_dirs) logs with verbosity on, but we don't have verbosity from config before reading the config file.. I think behavior should be the same for -v and `verbose = true`. 

* It seems `yes` was already implemented but the behavior is weird, the name is `assume_yes` and the config takes precedent to the command line? :confused:

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -4,8 +4,17 @@
 #    "~/.config/something"
 #]
 
-# Same options as the command line flag
+# Disable specific steps - same options as the command line flag
 #disable = ["system", "emacs"]
+
+# Run specific steps - same options as the command line flag
+#only = ["system", "emacs"]
+
+# Do not ask to retry failed steps (default: false)
+#no_retry = true
+
+# Run inside tmux
+#run_in_tmux = true
 
 # List of remote machines with Topgrade installed on them
 #remote_topgrades = ["toothless", "pi", "parnas"]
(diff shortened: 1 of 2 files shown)


config.example.toml
```diff
@@ -4,9 +4,15 @@
 #    "~/.config/something"
 #]
 
-# Same options as the command line flag
+# Disable specifi steps - same options as the command line flag
 #disable = ["system", "emacs"]
 
+# Run specific steps - same options as the command line flag
```
idanski: fix typo

config.example.toml
```diff
@@ -4,9 +4,15 @@
 #    "~/.config/something"
 #]
 
-# Same options as the command line flag
+# Disable specifi steps - same options as the command line flag
```
idanski: fix typo

r-darwish: So fix it 🙂

config.example.toml
r-darwish: I think we're missing `run_in_tmux`

PR #239, opened 2019-10-11
Handle gracefully config path not existing (Fixes #238)

### LINKED ISSUE #238: `.config` doesn't always exist
In fresh WSL `topgrade` fails with:
```
ERROR topgrade::config > Unable to write the example configuration file to /home/mateusz/.config/topgrade.toml: No such file or directory (os error 2)
Error: Error reading the configuration
Caused by: No such file or directory (os error 2)
```
This is caused by non existent `~/.config`, creating this directory fixes the issue but I think `topgrade` could print better error or create directory itself.

idanski:
Note: removed verbosity from config file, as failing while reading the config file needs a logger output and env_logger can't be init twice or changed after init as far as I could tell.

@r-darwish WDYT?

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -70,7 +70,7 @@ pub struct ConfigFile {
 
 impl ConfigFile {
     fn ensure(base_dirs: &BaseDirs) -> Result<PathBuf, Error> {
-        let config_path = base_dirs.config_dir().join("/plooopy/topgrade.toml");
+        let config_path = base_dirs.config_dir().join("topgrade.toml");
         if !config_path.exists() {
             write(&config_path, include_str!("../config.example.toml"))
                 .map_err(|e| {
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -64,19 +64,18 @@ pub struct ConfigFile {
     yay_arguments: Option<String>,
     no_retry: Option<bool>,
     run_in_tmux: Option<bool>,
-    verbose: Option<bool>,
```
r-darwish: We don't have to get rid of this. We could just perform two ifs, one before the configuration is loaded, which checks `opt.verbose`, and one after that checks `config.verbose`

src/config.rs
```diff
@@ -120,6 +119,26 @@ impl ConfigFile {
             .context(ErrorKind::Configuration)?;
         Ok(())
     }
+
+    /// Creates a new, blank config file.
```
r-darwish: We can probably derive `Default` instead of implementing a method.

src/config.rs
```diff
@@ -64,19 +64,18 @@ pub struct ConfigFile {
...
     run_in_tmux: Option<bool>,
-    verbose: Option<bool>,
     cleanup: Option<bool>,
     only: Option<Vec<Step>>,
 }
 
 impl ConfigFile {
     fn ensure(base_dirs: &BaseDirs) -> Result<PathBuf, Error> {
-        let config_path = base_dirs.config_dir().join("topgrade.toml");
+        let config_path = base_dirs.config_dir().join("/plooopy/topgrade.toml");
```
r-darwish: I suppose you did that for debugging

src/config.rs
idanski: hmm, but I want to call `builder.init()` before running config `load`, after running init() is there any way to update the logger?

idanski: I think the answer is no, unless you have another idea..?

https://github.com/seanmonstar/pretty-env-logger/issues/27

r-darwish: Yeah, you're correct.

PR #245, opened 2019-10-30
Fix whitespace in emacs lisp code

mdboom:
On emacs 26.2, when the `--eval` argument to emacs contains spaces, it results in this error: 

```
End of file during parsing
```

It is enough to reproduce this with:

```
emacs --batch --eval "(if 1 0 42)"
```

and likewise with escaping (not that it should matter):

```
emacs --batch --eval "(if\ 1\ 0\ 42)"
```

It does seem that replacing these with 0xA0 (non-breaking space) works, though, and it's the best workaround I could find, though hackish.

### DIFF
--- a/src/steps/emacs.rs
+++ b/src/steps/emacs.rs
@@ -41,11 +41,18 @@ impl Emacs {
 
         print_separator("Emacs");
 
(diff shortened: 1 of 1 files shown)


r-darwish:
Seems so. Thank you

PR #253, opened 2019-11-07
Extend support for Nix/NixOS/home-manager

filalex77:
This change is [Reviewable](https://reviewable.io/reviews/r-darwish/topgrade/253)

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -53,6 +53,7 @@ pub fn run_homebrew(cleanup: bool, run_type: RunType) -> Result<(), Error> {
 #[must_use]
 pub fn run_nix(run_type: RunType) -> Result<(), Error> {
     let nix = require("nix")?;
+    let nix_channel = require("nix-channel")?;
     let nix_env = require("nix-env")?;
 
     print_separator("Nix");
@@ -61,6 +62,7 @@ pub fn run_nix(run_type: RunType) -> Result<(), Error> {
     } else {
         run_type.execute(&nix).arg("upgrade-nix").check_run()?;
     }
+    run_type.execute(&nix_channel).arg("--update").check_run()?;
     run_type.execute(&nix_env).arg("--upgrade").check_run()
 }
 
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -64,6 +64,13 @@ pub fn run_nix(run_type: RunType) -> Result<(), Error> {
     run_type.execute(&nix_env).arg("--upgrade").check_run()
 }
 
+pub fn run_home_manager(run_type: RunType) -> Result<(), Error> {
+    let home_manager = require("home-manager")?;
+
+    print_separator("home-manager");
+    run_type.execute(&home_manager).arg("switch").check_run()
+}
+
 pub fn run_pearl(run_type: RunType) -> Result<(), Error> {
     let pearl = require("pearl")?;
     print_separator("pearl");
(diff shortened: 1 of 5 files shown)


src/steps/os/unix.rs
```diff
@@ -1,6 +1,7 @@
+use super::linux::Distribution;
```
r-darwish: This fails on macOS. You need to put `#cfg[unix]` before it.

src/steps/os/unix.rs
```diff
@@ -52,13 +53,26 @@ pub fn run_homebrew(cleanup: bool, run_type: RunType) -> Result<(), Error> {
 #[must_use]
 pub fn run_nix(run_type: RunType) -> Result<(), Error> {
     let nix = require("nix")?;
+    let nix_channel = require("nix-channel")?;
     let nix_env = require("nix-env")?;
 
     print_separator("Nix");
-    run_type.execute(&nix).arg("upgrade-nix").check_run()?;
+    if let Ok(Distribution::NixOS) = Distribution::detect() {
+        print_warning("Nix on NixOS must be upgraded via 'nixos-rebuild', skipping.");
```
r-darwish: I rather that this will just be skipped without a warning. You can add a debug log if you like, however.

src/steps/os/mod.rs
```diff
@@ -2,7 +2,7 @@
 pub mod dragonfly;
 #[cfg(target_os = "freebsd")]
 pub mod freebsd;
-#[cfg(target_os = "linux")]
+#[cfg(unix)]
```
r-darwish: I rather that we don't do this. Instead, the `nix` step should use that module only in Linux

filalex77: Got it. Thanks!

andrei-pavel:
Add `--non-interactive` to `nix-env` for `--yes` flag?

filalex77:
@andrei-pavel Not sure on this flag - can't find it in the manual. Can you please link me the relevant documentation? I would love to have it. Thanks!

src/steps/os/unix.rs
```diff
@@ -49,16 +49,49 @@ pub fn run_homebrew(cleanup: bool, run_type: RunType) -> Result<(), Error> {
     Ok(())
 }
 
+#[cfg(target_os = "linux")]
+#[must_use]
+pub fn run_nix(run_type: RunType) -> Result<(), Error> {
```
r-darwish: Isn't this step supposed to run on macOS? What I meant is that only the distribution checking will be performed on Linux. I also rather that the step will be entirely skipped without printing a seperator on nixOS.

You can achieve this by removing `[cfg(target_os = "linux")]` from the function, and instead move the if statement to line 59, before the separator. It should be something like this:

```rust
#[cfg(target_os = "linux")]
{
    if let Ok(Distribution::NixOs) ... {
        debug!("Skipping ...");
        return Err(ErrorKind::SkipStep.into());
    }
}

print_separator("Nix");
# proceed as usual
```

src/steps/os/unix.rs
```diff
@@ -49,16 +49,49 @@ pub fn run_homebrew(cleanup: bool, run_type: RunType) -> Result<(), Error> {
...
+    } else {
+        run_type.execute(&nix).arg("upgrade-nix").check_run()?;
+    }
+
+    Ok(())
+}
+
+#[cfg(not(target_os = "linux"))]
 #[must_use]
 pub fn run_nix(run_type: RunType) -> Result<(), Error> {
```
r-darwish: Isn't `run_nix` defined twice now?

src/steps/os/unix.rs
filalex77: Ah, `SkipStep`. Thanks for clarifying!

r-darwish:
Awesome. Thank you!

PR #259, opened 2019-11-19
[📦 NEW] PlugClean for vim on cleanup

andrei-pavel:


r-darwish:
Thanks. Can you please fix the failing tests?

andrei-pavel:
Fails on Windows, because cleanup is disabled.
### https://github.com/r-darwish/topgrade/blob/f506d523094317cffc8c93f6bae1e967c17af8d3/src/config.rs#L263-L264

So choices are do we enable cleanup for Windows or do we disable vim for Windows?

r-darwish:
It still fails in Windows because the `cleanup` method was never used there. Please remove [this line](https://github.com/r-darwish/topgrade/blob/f506d523094317cffc8c93f6bae1e967c17af8d3/src/config.rs#L263)

PR #260, opened 2019-11-19
[📦 NEW] flutter upgrade

andrei-pavel:


r-darwish:
Can you also check if the directory that should contain the binaries exists? Like we [do with Gem](https://github.com/r-darwish/topgrade/blob/465c32abab93113c9aca491de4a7e5d524e8f0f8/src/steps/generic.rs#L23)

src/steps/generic.rs
```diff
@@ -18,6 +18,13 @@ pub fn run_cargo_update(run_type: RunType) -> Result<(), Error> {
         .check_run()
 }
 
+pub fn run_flutter_upgrade(run_type: RunType) -> Result<(), Error> {
+    let flutter = utils::require("flutter")?;
+    print_separator("Flutter");
+
```
r-darwish: Can you also check if the directory that should contain the binaries exists? Like we do with Gem

andrei-pavel: I find this difficult. Flutter SDK doesn't have a default location and can't be deduced from a dedicated environment variable. The way I install it is download an archive from their website, extract it to `/opt/flutter` and add `/opt/flutter/bin` to my `${PATH}` in `~/.profile`. [AUR](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=flutter) installs it to `/opt/flutter` as well, but it can't be taken for granted on other systems. But `utils::require("flutter")?` does check for the binary's existence, right? And the SDK that is updated is right next door, one directory level upwards, so it could be assumed that it exists? :man_shrugging:

r-darwish: Well, I guess we can pass on this. `Flutter` isn't a common thing anyway.

PR #261, opened 2019-11-19
[📦 NEW] go get -u all

andrei-pavel:


src/steps/generic.rs
```diff
@@ -18,6 +18,13 @@ pub fn run_cargo_update(run_type: RunType) -> Result<(), Error> {
         .check_run()
 }
 
+pub fn run_go(run_type: RunType) -> Result<(), Error> {
+    let go = utils::require("go")?;
+    print_separator("Go");
```
r-darwish: Can you please check that the binary directory exists [like we do with Gem](https://github.com/r-darwish/topgrade/blob/465c32abab93113c9aca491de4a7e5d524e8f0f8/src/steps/generic.rs#L23)? Try to get the value of `GOPATH` or the default directory if this variable doesn't exist.

r-darwish:
Looking good. Can you please resolve the conflicts?

PR #265, opened 2019-11-26
Add zplugin to zsh steps

kmcopper:
This pull request adds additional plugin manager to the zsh steps.

### DIFF
--- a/src/steps/zsh.rs
+++ b/src/steps/zsh.rs
@@ -55,6 +55,24 @@ pub fn run_zplug(base_dirs: &BaseDirs, run_type: RunType) -> Result<(), Error> {
     run_type.execute(zsh).args(&["-c", cmd.as_str()]).check_run()
 }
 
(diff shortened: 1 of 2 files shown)


src/steps/zsh.rs
```diff
@@ -55,6 +55,24 @@ pub fn run_zplug(base_dirs: &BaseDirs, run_type: RunType) -> Result<(), Error> {
...
+
+    env::var("ZPLGM[HOME_DIR]")
+        .map(PathBuf::from)
+        .unwrap_or_else(|_| base_dirs.home_dir().join(".zplugin"))
+        .require()?;
+
+    print_separator("zplugin");
+
+    let cmd = format!(
+        "aource {} && zplugin self-update && zplugin update --all",
```
r-darwish: You probably meant `source`

kmcopper: Oops! Thanks for catching this.  Probably happened when I was hastily fixing it in vim.

r-darwish:
Thanks

PR #280, opened 2019-12-20
Support updating asdf plugins

liaden:
There can be a lot of them: https://github.com/asdf-vm/asdf-plugins

I was thinking about the updating for the underlying versions of things that are managed via asdf, but that has a lot more potential complexity there.

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -172,6 +172,15 @@ pub fn run_custom_command(name: &str, command: &str, ctx: &ExecutionContext) -> 
     ctx.run_type().execute(shell()).arg("-c").arg(command).check_run()
 }
 
+#[cfg(unix)]
+pub fn run_asdf_update(run_type: RunType) -> Result<()> {
+    let asdf = utils::require("asdf")?;
+
+    print_separator("Asdf");
+
+    run_type.execute(&asdf).args(&["plugin-update", "--all"]).check_run()
+}
+
 pub fn run_composer_update(base_dirs: &BaseDirs, run_type: RunType) -> Result<()> {
     let composer = utils::require("composer")?;
     let composer_home = Command::new(&composer)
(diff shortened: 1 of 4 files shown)


src/steps/generic.rs
```diff
@@ -171,6 +171,11 @@ pub fn run_custom_command(name: &str, command: &str, run_type: RunType) -> Resul
     run_type.execute(shell()).arg("-c").arg(command).check_run()
 }
 
+pub fn run_asdf_update(run_type: RunType) -> Result<()> {
+    let asdf = utils::require("asdf")?;
+    run_type.execute(&asdf).args(&["plugin-update", "--all"]).check_run()
```
r-darwish: Add a separator printing

src/steps/generic.rs
```diff
@@ -171,6 +171,11 @@ pub fn run_custom_command(name: &str, command: &str, run_type: RunType) -> Resul
     run_type.execute(shell()).arg("-c").arg(command).check_run()
 }
 
+pub fn run_asdf_update(run_type: RunType) -> Result<()> {
```
r-darwish: I don't think asdf supports windows, so you should wrap this with `#[cfg(unix)]`

liaden: I followed the example as from tmux. I'm unsure if that is the correct approach, or if I should just annotate the whole method with the `#[cfg(unix)]`.

KokaKiwi: Personally i would put the whole method under `#[cfg(unix)]` because the step will be run anyway as it's listed in the `Step` enum:
https://github.com/r-darwish/topgrade/blob/def4c3500c96c7761bf61e61631ae882184a7fb7/src/config.rs#L256
Which would actually lead to topgrade just crashing on Windows because it would try to run the step and trigger the `unreachable!` (which i think is not the correct macro to use too for this kind of error)

r-darwish:
@liaden could you please merge this branch with the master branch and apply the required changes?

PR #294, opened 2020-01-09
Fix tldr upgrade when using tealdeer

severinkaderli:
I'm using [tealdeer](https://github.com/dbrgn/tealdeer) as my tldr client and when using topgrade the ugprade didn't work because tealdeer only executed the -v command and not the --update one. This PR changes the step to also work with tealdeer.


r-darwish:
`-v` isn't supposed to be a command, but instead cause TLDR to print a message while upgrading. I don't mind just removing `-v`.

severinkaderli:
I went ahead and removed the -v call of TLDR. My TLDR client actually outputs a message when updating:
image

PR #312, opened 2020-01-30
Add support for Fedora Silverblue

K-Finlay:
Fedora Silverblue doesn't use dnf or yum for package management, rather, it uses rpm-ostree.
This pull request adds support for rpm-ostree.


r-darwish:
Thanks

PR #319, opened 2020-02-10
Support update of zsh plugin with antibody

mhaessig:
This PR adds support for updating zsh plugins using [antibody](https://getantibody.github.io/).


r-darwish:
Thanks

PR #320, opened 2020-02-12
zplugin was renamed to zinit

hugoh:
zplugin was renamed to zinit: https://github.com/zdharma/zinit/commit/1ca8f6856cfe5e56abc969eae8e2a5489feb3a65

* s/zplugin/zinit/
* Testing for $ZPFX instead of ZPLGM which is exported

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -217,7 +217,7 @@ fn run() -> Result<()> {
             runner.execute("antibody", || zsh::run_antibody(run_type))?;
             runner.execute("antigen", || zsh::run_antigen(&base_dirs, run_type))?;
             runner.execute("zplug", || zsh::run_zplug(&base_dirs, run_type))?;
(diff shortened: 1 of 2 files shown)


hugoh:
Tests are failing in files I did not touch, so I'll leave it there for now.

r-darwish:
Thanks

strangelittlemonkey:
I know this has already been merged but I’m not sure this is the best way to handle it. Checking their page it looks like the rename happened less than a month ago. What happens for the people that have it and it’s called zplugin and isn’t the new project name yet. It might be better to check for either and give one precedence over the other rather than to totally exclude the name, at least this early in a project renaming.

hugoh:
That's a good point that both should really be checked.

hugoh:
@strangelittlemonkey - I attempted to address this in #321

PR #322, opened 2020-02-16
[🐛 FIX] ignore EXIT_NOTHING_TO_DO for all fwupdmgr commands

### LINKED ISSUE #246: Firmware updates report failed status when no updates found
The firmware (fwupd) update step has started failing for me when there are no updates. I run arch linux. If there is any debug info I can provide please let me know.

andrei-pavel:
Fixes #246.

The issue is detailed in [1] and following comments.

Based on the same reasoning as for "get-updates" discussed in [2], return code 2 (EXIT_NOTHING_TO_DO) would need to be ignore for "refresh" as well.

[1] https://github.com/r-darwish/topgrade/issues/246#issuecomment-586640775-permalink
[2] https://github.com/r-darwish/topgrade/issues/246#issuecomment-549117042-permalink

### DIFF
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -403,12 +403,13 @@ pub fn run_fwupdmgr(run_type: RunType) -> Result<()> {
 
     print_separator("Firmware upgrades");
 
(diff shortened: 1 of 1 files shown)


r-darwish:
Thank you

PR #324, opened 2020-02-16
Comply with new xbps behaviour

slotThe:
As per void-linux/xbps#194, `xbps-install` now forces one to first sync and upgrade `xbps` itself before doing a full system update.


r-darwish:
Thanks

PR #341, opened 2020-02-26
MacPorts (fix #331)

### LINKED ISSUE #331: Add MacPorts support
If you aim for completeness, supporting MacPorts (both as a) an install source for topgrade and b) as a supported package manager) would be great.

I'm happy to create a topgrade port (a) but since my rust knowledge is non existing, someone else would have to make topgrade call something like 
`sudo port selfupdate && sudo port -u upgrade outdated && sudo port -N reclaim`
if macports is installed. This should be easy to detect by checking running 
`which port`

r-darwish:


r-darwish:
@lhaeger please test

lhaeger:
> @lhaeger please test

Seems to work:

```
―― 10:14:55 - MacPorts ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――
--->  Updating MacPorts base sources using rsync
MacPorts base version 2.6.2 installed,
MacPorts base version 2.6.2 downloaded.
--->  Updating the ports tree
--->  MacPorts base is already the latest version

The ports tree has been updated. To upgrade your installed ports, you should run
  port upgrade outdated
Nothing to upgrade.
--->  Scanning binaries for linking errors
--->  No broken files found.
--->  No broken ports found.
```
but it cannot be run exclusively nor be disabled:

```
./topgrade --only macports -v                                                                                                                                                                                  
error: 'macports' isn't a valid value for '--only <only>...'
	[possible values: atom, cargo, composer, emacs, firmware, flutter, gem, git_repos, go, jetpack, myrepos, node, opam, package_managers, pearl, pipx, remotes, restarts, rustup, sdkman, shell, stack, system, tldr, tlmgr, vcpkg, vim] 
```

Could you add "macports" (and probably also "brew") to the list?

r-darwish:
Both brew and macports can be disabled using `package_managers`

Issue #337, opened 2020-02-23
Retry after password timeout


domoaligato:
**What did you expect to happen?**
When I type my password the first time it should update all packages for all package managers using that password without re-prompting for my password.

**What actually happened?**
I type my password and then confirm the pacman updates I wish to install. If there are a lot of them then it will ask me for my password when it gets to the installation part of the aur updates. I have stepped away from my system for a few minutes and when I return I find that instead of topgrade just reusing the password that I already provided. It tells me that it timed out waiting for password input and asks me to retry. When I retry it is retrying the entire topgrade process. Not where it left off at in the aur upgrade process (normally this is after the compile when it is actually doing stuff that needs sudo). I believe that it is just hitting the sudo timeout. I do not want to change my global sudo timeout and am wondering if topgrade could keep it alive somehow in the background while it is running?

**Additional details:**
- Which operating system or Linux distribution are you using? Arch
- How did you install Topgrade? AUR

<details><pre>
Paste the output of `topgrade -v`
</pre></details>

Linux4Life531:
How long does it take it to install the parts before it asks you to enter the password for the second time?
If you are the SuperUser of the system or you know the password, entering sudo su and it will enter you into sudo mode.
Then enter topgrade and it will automatically enter the password for you.

The reason why what you are doing is not working because the auto-enter-sudo-password is timing out. If you do not want to change the time, then my above option is the best. Good Luck!

Linux4Life531:
Oh and hi again!
I am also trying to solve your man page issue which I have put in a pull request for.

r-darwish:
@domoaligato I wish we had some solution for this. I'm too annoyed by it. 

There are probably some potential hacks to achieve this like spawning a process in the beginning that can be acted as a "launcher". However, sudo is a sensitive thing so I don't want to implement risky solutions. 

I'll keep this open in case someone that understands sudo better than me comes up with some risk-less solution

domoaligato:
the end game is that I wish to use topgrade unattended. That is why I asked about a man page. I thought that maybe I could accept all aur and pacman updates automatically without prompts (another feature request). sorry for the spam. I just thought that things like this would need attention before the project would be at that point. Thank you for your wonderful work.

[...]

PR #346, opened 2020-02-27
Add an option for self-updating composer (fix #318)

### LINKED ISSUE #318: Composer should uprade itself
I just noticed that my composer executable was outdated many months and I got an email about Github API deprecated use. That was fixed in the last composer release https://github.com/composer/composer/releases/tag/1.9.3 , but `topgrade` does not upgrade composer itself. I think it should?

A recommended global installation of composer is in /usr/local/bin on Linux/MacOs/Unix, see https://getcomposer.org/doc/00-intro.md#globally . Composer has a built in slef-update command, so it should be enough to run `sudo composer self-update`.

r-darwish:


r-darwish:
@klausi please test this change by putting the following snippet in your `topgrade.toml`:

```toml
[composer]
self_update = true
```

klausi:
Thanks, it works with 2 minor problems:

1. It asks you for sudo again. I already typed in my sudo password at the System update (APT) part. Topgrade should remember the sudo password for 5 minutes and not ask for it twice in 1 command run. Maybe that is a separate issue?
2. There is no output:
```
―― 13:52:29 - Composer ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――
[sudo] password for klausi: 

―― 13:52:35 - Firmware upgrades ――――――――――――――――――――――――――――――――――――――――――――――――
...
```
Did it upgrade Composer or not? Was it successful or not? Maybe that is also by design?

r-darwish:
1. We're trying to think about a solution in #337. Sadly, Topgrade doesn't control the sudo timeout.
2. I pushed a fix for that. Can you please pull and check again?

klausi:
Thank you, output looking good now!

I see, so the sudo thing is a bit annoying. Can you prevent asking for sudo if there is no global composer update anyway? If current composer version is latest version do not ask for sudo.

r-darwish:
@klausi is there any way to check it by running composer?

klausi:
You could run `composer self-update` without sudo. If exit status is 0 then it means there is no update. Otherwise there is an error, probably `rename(/usr/local/bin/composer): failed to open stream: Permission denied` and the exit status is 1. Then you could re-run `self-update` with sudo.

r-darwish:
I implemented your suggestion. Please check.

klausi:
Works for me, thanks!

The git commits are a bit confusing to review here, looks like you mixed in some unrelated changes?

r-darwish:
I merged it with the master branch. I perform squash merges so it will be a single commit when merged. I suggest that you look at the changes tab instead of the commit tab

klausi:
cool, I like squash merges :-)

You probably did not merge in the recent master branch changes into this branch, that's why I see lots of unrelated changes at the changes tab.

PR #349, opened 2020-02-29
Improve man page formatting

idanski:
Before:
image

After:
Screenshot_20200229_140341


Linux4Life531:
Brilliant, thanks for the help as the first page I made was not very good. It was just to fix the bug.

r-darwish:
Thanks @idanski

Issue #285, opened 2020-01-02
Desktop notifications


r-darwish:

Linux4Life531:
Has this issue been fixed or not. If bot, I might be able to look into it and try to find a solution.

r-darwish:
It's currently implemented only in macOS

Linux4Life531:
I will have a look into it for linux then.

Linux4Life531:
What is the last program that runs on the Linux version of topgrade?

r-darwish:
We should run [notify-send](https://ss64.com/bash/notify-send.html) in Linux instead of relying `notify-rust`, as the former adds a link with a native library.

thomas-merz:
@r-darwish - Regarding your commit https://github.com/r-darwish/topgrade/commit/b0491519930029aadb1ee2f269120fd059e2b621: is or will be there also a parameter to **disable** Desktop Notifications on MacOS? 🤔

PR #350, opened 2020-02-29
Add notification display for linux targets

marty-oehme:
## Standards checklist:

- [ ] The code compiles
- [x] The code passes rustfmt
- [x] The code passes clippy
- [x] The code passes tests

Adds notifications for linux using the same crate as macos, to help with #285 

The only difference is that the dbus notification takes timeouts in ms instead of s, so `.timeout(5)` instantly made the notification disappear again. I set it to 5000 to mirror 5 seconds, though at some point a configurable delay might be preferable?

I am not sure if the way I am including the dependency in Cargo.toml is the best; if there's a better way let me know and I will fix it!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,9 @@ rust-ini = "0.14.0"
 [target.'cfg(target_os = "macos")'.dependencies]
 notify-rust = "3.6.3"
 
+[target.'cfg(target_os = "linux")'.dependencies]
+notify-rust = "3.6.3"
+
 [target.'cfg(unix)'.dependencies]
 nix = "0.17.0"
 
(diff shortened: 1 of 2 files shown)


marty-oehme:
Ah, building on ubuntu fails at running the build command for libdbus-sys v0.2.1 unless the `libdbus-glib-1-dev` package is available.

r-darwish:
@marty-oehme yeah, this is why I didn't enable it on Linux in the first place. I think that the best strategy would be not to use this library in Linux, but instead look for a command line tool that can send notifications are run it instead.

marty-oehme:
@r-darwish That's completely fair, I didn't know the history of the issue here sorry! 

I think relying on a cli like libnotify or [toastify](https://crates.rs/crates/toastify) is actually a good workaround.

I'll close the pr for now.

PR #351, opened 2020-02-29
Fix Go packages when using Go modules

r-darwish:


fydai:
Shouldn't the environment variable be `GO111MODULE` and not `GO111MODLE`?

r-darwish:
@fydai good catch, thanks. I fixed it.

PR #353, opened 2020-02-29
Desktop notifications for Linux

r-darwish:


r-darwish:
@idanski - can you please test this change in Linux? I don't have GUI in my VM.

idanski:
@r-darwish 
Sure. I'll try to do it later today/tomorrow.

Thanks!

src/terminal.rs
```diff
@@ -59,23 +69,41 @@ impl Terminal {
...
+                if let Some(timeout) = timeout {
+                    notification.timeout(Timeout::Milliseconds(timeout.as_millis() as u32));
+                }
+                notification.show().ok();
+            } else if #[cfg(target_os = "linux")] {
+                if let Some(ns) = self.notify_send.as_ref() {
+                    let mut command = Command::new(ns);
+                    if let Some(timeout) = timeout {
+                        command.arg("-t");
+                        command.arg(format!("{}", timeout.as_millis()));
```
idanski: I think you forgot to add the message :)

idanski:
Screenshot_20200303_132030

@r-darwish it works but I suggest adding param `-a Topgrade` to have a good title instead of the default notify-send.

PR #357, opened 2020-03-03
Avoid searching for git repos inside already detected repos

r-darwish:


r-darwish:
@NerdyMcNerd can you please [test this change](https://github.com/r-darwish/topgrade/wiki/Running-Topgrade-from-source)?

You can run topgrade with `--v` to see if it skips detection of sub repositories

NerdyMcNerd:
Now I have a working build environment (which is pretty streight forward compared to C/C++).
I got the same compiler error as the CI Build, but changing `main.rs` (line 198) to use `git_repos.insert_if_repo`instead of `git_repos.insert` makes it work.

The test on my Work-PC (with 44 Git Repos) revealed:
- The latest release version of topgrade was stuck scanning for Git Repos ~5min 
- This Branch's (debug) version only 14sec.  
(Verbose mode took longer due to all the output logging, but most of it was "skipping" messages)

--> Big improvement

- The release version detected all submodules as own Git repos and tried to pull them (and failed)
- This Branch's version, pulled only the top level git repos

--> is good for me, since submodules should be handled by git internally instead of being pulled by topgrade

When you fix the compiler error in `main.rs`on line 198, I see this as fixed.

PR #366, opened 2020-03-10
added rhel to CentOS

ImNtReal:
I'm not sure if this is how you'd like to handle support RHEL, but it seems to work for me, so far.

## Standards checklist:

- [x] The code compiles
- [x] The code passes rustfmt
- [x] The code passes clippy
- [x] The code passes tests

### DIFF
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -53,7 +53,7 @@ impl Distribution {
         }
 
         Ok(match id {
-            Some("centos") | Some("ol") => Distribution::CentOS,
+            Some("centos") | Some("rhel") | Some("ol") => Distribution::CentOS,
             Some("clear-linux-os") => Distribution::ClearLinux,
             Some("fedora") => Distribution::Fedora,
             Some("void") => Distribution::Void,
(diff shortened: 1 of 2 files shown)


ImNtReal:
Forgive me for not having much experience with PRs in GitHub, but after creating a new commit that includes the os-release file, do I need to do anything to include that commit in the PR?

Also, I was wondering if there was really any difference between what topgrade does when it detects CentOS vs Fedora. Maybe they should be merged with RHEL into one OS type of RedHat, so that finding any of those strings in the os-release file will cause it to follow one path from there.

r-darwish:
As you can see [here](https://github.com/r-darwish/topgrade/blob/f9374e3ddf0acc232ed5fa96b70872b562001997/src/steps/os/linux.rs#L88), Fedora and RHEL based systems actually run the same function. Can you please just add another test with your new file below [this one](https://github.com/r-darwish/topgrade/blob/f9374e3ddf0acc232ed5fa96b70872b562001997/src/steps/os/linux.rs#L480-L484)?

r-darwish:
Thank you

PR #370, opened 2020-03-14
Look for Emacs directory in $XDG_CONFIG_HOME

slotThe:
## Standards checklist:

- [X] The PR title is descriptive.
- [X] The code compiles
- [x] The code passes rustfmt
- [X] The code passes clippy
- [X] The code passes tests
- [X] *Optional:* I have tested the code myself

Starting with Emacs 27, Emacs now looks for an `emacs` directory in `$XDG_CONFIG_HOME` (c.f. [this commit](
https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=4118297ae2fab4886b20d193ba511a229637aea3)) if no `~/.emacs.d` directory is present, so we better do that as well.

### DIFF
--- a/src/steps/emacs.rs
+++ b/src/steps/emacs.rs
@@ -3,7 +3,7 @@ use crate::terminal::print_separator;
 use crate::utils::{require, require_option, PathExt};
 use anyhow::Result;
 use directories::BaseDirs;
-#[cfg(windows)]
+#[cfg(any(windows, target_os = "macos"))]
 use std::env;
 use std::path::{Path, PathBuf};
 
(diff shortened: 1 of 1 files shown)


src/steps/emacs.rs
```diff
@@ -21,7 +21,11 @@ pub struct Emacs {
 impl Emacs {
     fn directory_path(base_dirs: &BaseDirs) -> Option<PathBuf> {
         #[cfg(unix)]
-        return base_dirs.home_dir().join(".emacs.d").if_exists();
+        return base_dirs
+            .home_dir()
+            .join(".emacs.d")
+            .if_exists()
+            .or_else(|| base_dirs.config_dir().join("emacs").if_exists());
```
r-darwish: `base_dirs.config_dir()` returns `~/Library/Preferences` in macOS. I'm not sure this is where Emacs 27 expects to find its configuration in macOS.

slotThe: Ah indeed, looking at the code emacs checks for the existence of the `$XDG_CONFIG_HOME` environment variable and, in case it doesn't exist, falls back to `~/.config/` (which would probably be the case for macOS).  Apologies for that.

I've force pushed a change that should sort this out (although this feels like a hack to me, so feel free to suggest something more idiomatic).

r-darwish: Seems reasonable, but you should also check if the variable `XDG_CONFIG_HOME` itself exists.

slotThe: You mean on macOS?  Right, I thought these env variables were just not a thing on macOS...

I've pushed something that should (hopefully) work

r-darwish:
Looking good. Can you please just fix the Clippy errors shown in Github actions?

slotThe:
Alright, should be all done.

r-darwish:
Thanks

PR #375, opened 2020-03-17
Update pihole only if using the pihole distribution (fix #373)

### LINKED ISSUE #373: Option to ignore Pi-hole
I am running Pi-hole on Arch Linux using the **pi-hole-server** AUR package.

As the package takes care of updating Pi-hole itself, the "-up" switch has been removed and throws an error message when topgrade tries to run the update command.

Topgrade should have an option to ignore Pi-hole, e.g. `disable = ["pihole"]`

r-darwish:


r-darwish:
@xyrolaith can you please test this change?

xyrolaith:
Looks fine to me. Ran the compile with the fix on one of my systems and Pi-hole update has been skipped as expected.

PR #389, opened 2020-04-03
Disable tlmgr in linux (fix #384)

### LINKED ISSUE #384: Disable tlmgr by default
**I want to suggest some general feature**
I think we should disable `tlmgr` by default.  
The background is, that updating with `tlmgr`, if texlive is installed via your package manager, breaks your existing texlive installation. But at least on Fedora texlive also installs `tlmgr` by default.

This is the message which always gets printed during the `tlmgr` update:
```
―― 23:06:07 - TeX Live package manager ―――――――――――――――――――――――――――――――――――――――――
*** WARNING ***: Performing this action will likely destroy the Fedora TeXLive install on your system.
*** WARNING ***: This is almost NEVER what you want to do.
*** WARNING ***: Try using dnf install/update instead.
*** WARNING ***: If performing this action is really what you want to do, pass the "ignore-warning" option.
*** WARNING ***: But please do not file any bugs with the OS Vendor.
```

While it's not dangerous per se, it looks very dangerous and therefore does not really give a good user experience.

r-darwish:


Evpok:
Sorry to bother you, but is it possible to enable it back using the configuration? The example shows how to disable things but not how to re-enable them (short of using a custom command, that is).

PR #432, opened 2020-06-15
reorg linux os check, add ID_LIKE=fedora, add tests

chhe:
- check for specific ID first, then check for ID_LIKE
- added case for ID_LIKE=fedora
- added test for Fedora Remix on WSL
- added test for Pengwin (Debian-like on WSL)

## Standards checklist:

- [x] The code compiles
- [x] The code passes rustfmt
- [x] The code passes clippy
- [x] The code passes tests
- [x] I have tested the code myself

I hope this PR is better than my last. Regarding the reordering of the OS checks: it makes more sense to me to check for a specific ID first and then check the ID_LIKE cases. The tests run. I hope you don't mind. I'm still very new to Rust.


r-darwish:
Thanks!

PR #446, opened 2020-06-24
Don't run zsh in interactive mode to determine zsh_custom

r-darwish:


r-darwish:
@ffloimair can you please test this change?

ffloimair:
I have no experience with rust yet.
I have rust installed via homebrew on MacOS and tried "cargo build" in my local git copy, but unfortunately I get compile errors while compiling "mac-notification-sys- v0.3.0"

If you give me a heads up what to do exactly to get this built, I can give it another try.

r-darwish:
I suppose your probably need xcode or something. I think I'll merge this change and we'll hope for the best.

PR #451, opened 2020-06-26
Use a vim script to upgrade vim

r-darwish:


r-darwish:
@openjck @cujomalainey @KokaKiwi - I tried creating a vim script that detects which plugin system you're using and use the appropriate one. However, this doesn't seem to work as expected. I see the `echo "one"` printing despite the fact that there's no such command as `PlugUpgrade1`, and the other echos do not appear. Does anyone knows enough vimscript to help with this PR?

openjck:
I believe vimscript is like Bash in that execution continues even after an error is encountered. In this case, I believe line 1 raises an error but line 2 echoes "one" regardless.

openjck:
I wonder if the `exists` calls are failing due to [this caveat](https://stackoverflow.com/a/5010399/4297741).

cujomalainey:
Sorry I'm not a vim script expert.

r-darwish:
I think I got it working. Can any of you please test this branch?

openjck:
I'll take a look today or tomorrow. Thank you!

openjck:
I'm sorry, I'm not very familiar with Rust. If you get a chance, can you please share instructions for testing this locally?

openjck:
I ended up doing a `cargo install --path .`, it works well for me! :smiley:

PR #461, opened 2020-07-01
Add ability to ignore certain failures

r-darwish:


r-darwish:
@derphilipp can you please test the change?

derphilipp:
That is exactly what I wished for, awesome work!

Issue #380, opened 2020-03-21
executing topgrade on remote fails


Byter09:
**What did you expect to happen?**
Using the `remote_topgrades = ["btr-srv.lan"]` configuration option, I expect topgrade to connect to my server and run `topgrade` there as well.

**What actually happened?**
It successfully asks for my password to connect, but then says that `env` can't find `topgrade`.

**Additional details:**
- Which operating system or Linux distribution are you using?
  - Host: `Linux BTR-PC 5.5.8-1-MANJARO #1 SMP PREEMPT Thu Mar 5 20:29:51 UTC 2020 x86_64 GNU/Linux`
  - Server: `Linux BTR-SRV 5.5.9-2-MANJARO #1 SMP PREEMPT Sun Mar 15 10:32:03 UTC 2020 x86_64 GNU/Linux` (yes I run Manjaro on a server)
- How did you install Topgrade?
**Both** machines used
  - `cargo install`
  - updates using `cargo-update`
- If a specific step is failing, what happens when you run the command on your own? (Run `topgrade --dry-run` to see which commands Topgrade is running)
  - It returns the exact same error.

<details><pre>
 2020-03-21T18:41:38.982Z DEBUG topgrade::config > Configuration at /home/btr/.config/topgrade.toml
 2020-03-21T18:41:38.982Z DEBUG topgrade::config > Loaded configuration: ConfigFile { pre_commands: None, commands: None, git_repos: None, predefined_git_repos: None, disable: None, remote_topgrades: Some(["btr-srv.lan"]), ssh_arguments: None, git_arguments: None, tmux_arguments: None, set_title: None, assume_yes: None, yay_arguments: None, no_retry: None, run_in_tmux: None, cleanup: None, notify_each_step: Some(true), accept_all_windows_updates: None, only: None, composer: None }
 2020-03-21T18:41:38.982Z DEBUG topgrade::utils  > Detected "/usr/bin/notify-send" as "notify-send"
 2020-03-21T18:41:38.982Z DEBUG topgrade         > Version: 4.3.0
 2020-03-21T18:41:38.982Z DEBUG topgrade         > OS: x86_64-unknown-linux-gnu
 2020-03-21T18:41:38.982Z DEBUG topgrade         > Args { inner: ["topgrade", "-v"] }
 2020-03-21T18:41:38.982Z DEBUG topgrade         > Binary path: Ok("/home/btr/.cargo/bin/topgrade")
[...]

PR #466, opened 2020-07-08
Add remote_topgrade_path configuration option

Aehmlo:
This (slightly hacky) option provides a global method of specifying a Topgrade executable outside of the standard $PATH (see #380). For instance, to use Topgrade executables from ~/.cargo/bin, one may specify `remote_topgrade_path = ".cargo/bin/topgrade"`. The remote upgrade command will then be something along the lines of `ssh -t <host> env TOPGRADE_PREFIX=<host> .cargo/bin/topgrade`, which is then resolved relative to the user's home directory.

This implementation opts to make this a global setting rather than undertake per-remote configuration, which is likely to require significantly more effort. However, pursuing per-remote setup is likely
a worthy future goal. This implementation also supports only UTF-8 strings, i.e. std::string::String, which is internally consistent but necessarily misses some valid setups.

## Standards checklist:

- [x] The code compiles
- [x] The code passes rustfmt
- [x] The code passes clippy
- [x] The code passes tests

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -22,6 +22,9 @@
 # Arguments to pass SSH when upgrading remote systems
 #ssh_arguments = "-o ConnectTimeout=2"
 
+# Path to Topgrade executable on remote machines
+#remote_topgrade_path = [".cargo/bin/topgrade"]
+
 # Arguments to pass tmux when pulling Repositories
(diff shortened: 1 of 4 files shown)


r-darwish:
Thanks

PR #478, opened 2020-07-15
Modify powershell step to fix compilation on macOS

DivineGod:
I noticed macOS compilation failing on powershell after commit https://github.com/r-darwish/topgrade/commit/abc6a8065b5ff92d65c4eb2c21e0b95b937cf393

The change there to use `ExecutionContext` would fail because it was conditionally compiled in only on windows, but the update function would be used on all targets.

## Standards checklist:

- [x] The code compiles
- [x] The code passes rustfmt
- [x] The code passes clippy
- [x] The code passes tests


r-darwish:
Thanks

PR #487, opened 2020-07-29
Support sheldon (fix #484)

### LINKED ISSUE #484: Consider supporting sheldon
sheldon is a general purpose shell plugin manager

sheldon, https://github.com/rossmacarthur/sheldon

Generally all the ones topgrade currently supports

Check for sheldon in the path?

I don't think a lib is used for git in topgrade so I guess you'd check some env vars, find the correct base for repos, walk a tree looking for `.git` then `git pull`. Maybe just run `sheldon lock` either way.

**I want to suggest some general feature**

**More information**
I'm not a rust expert but have been trying to learn more and would be happy to make a PR if someone with commit powers could tell me that they would accept a particular functionality and OK a broad outline of implementation.

r-darwish:


r-darwish:
@xaocon can you please test this change?

xaocon:
```
―― 15:08:55 - Sheldon ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
Loaded ~/.zsh/sheldon/plugins.toml
   Checked https://github.com/DarrinTisdale/zsh-aliases-exa
   Checked https://github.com/xaocon/ssh
   Checked https://github.com/b4b4r07/enhancd
   Checked https://github.com/zdharma/fast-syntax-highlighting
   Checked https://github.com/zsh-users/zsh-completions
Locked ~/.zsh/sheldon/plugins.lock
```
Looks good to me!

PR #488, opened 2020-07-29
Skip needrestart on redhat based distributions (fix #485)

### LINKED ISSUE #485: Don't explicitly call needrestart
At least when using `dnf` on Fedora 32 (I have now clue how other package managers deal with that), `needrestart` is automatically triggered by `dnf` when it finished its transaction.  

However, due to `needrestart` being called explicitly at the last `topgrade` step, it is called twice. This means `needrestart` is also called every time even though there was no package change by `dnf`.

Therefore I would suggest to not call `needrestart` explicitly by `topgrade`.

r-darwish:


r-darwish:
@BjoernDaase - Can you please test this change?

BjoernDaase:
@r-darwish Works as expected, thank you!

FWIW, I would also skip the step on RHEL as it also uses dnf/yum.

r-darwish:
Topgrade doesn't support RHEL at the moment because nobody asked and I don't have an access to a RHEL instance to test it out. If RHEL will be supported in the future then I'll make sure needrestart is also skipped there.

PR #489, opened 2020-07-29
Don't pass --user-install if ruby is managed by rbenv (fix #393)

### LINKED ISSUE #393: Please support rbenv by making `--user-install` optional in the gem step
Since I use [rbenv](https://github.com/rbenv/rbenv) to manage my Ruby versions, gems will be installed "system wide" in my home folder, for example:

~/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0

However, since `topgrade` always uses the option `--user-install` when executing `gem update` it will install/update gems in:

~/.gem/ruby/2.7.0

Please make using that parameter `--user-install` optional (configurable via the config file) so that:
- topgrade will not install a duplicate set of gems
- topgrade will update the correct set of gems (those that I installed, not the duplicates from topgrade)

r-darwish:


r-darwish:
@pro-sumer can you please test this change?

pro-sumer:
> @pro-sumer can you please test this change?

Works for me!

PR #525, opened 2020-09-23
Add rtcl::rupdate

niheaven:
- R tools for the command line
- https://github.com/rdatsci/rtcl
- Like Jetpack, but more powerful.

**`rupdate`'s a sub command from `rtcl` package, and I don't know how to properly name it in topgrade**

## Standards checklist:

- [x] The code compiles
- [x] The code passes rustfmt
- [x] The code passes clippy
- [x] The code passes tests

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -280,6 +280,7 @@ fn run() -> Result<()> {
     runner.execute(Step::Krew, "krew", || generic::run_krew_upgrade(run_type))?;
     runner.execute(Step::Gem, "gem", || generic::run_gem(&base_dirs, run_type))?;
     runner.execute(Step::Sheldon, "sheldon", || generic::run_sheldon(&ctx))?;
+    runner.execute(Step::Rtcl, "rtcl", || generic::run_rtcl(&ctx))?;
 
     #[cfg(target_os = "linux")]
     {
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -124,7 +124,7 @@ pub fn run_jetpack(run_type: RunType) -> Result<()> {
 pub fn run_rtcl(ctx: &ExecutionContext) -> Result<()> {
     let rupdate = utils::require("rupdate")?;
 
-    print_separator("rtcl::rupdate");
+    print_separator("rtcl");
 
     ctx.run_type().execute(&rupdate).check_run()
 }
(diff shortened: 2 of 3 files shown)


src/main.rs
```diff
@@ -280,6 +280,7 @@ fn run() -> Result<()> {
     runner.execute(Step::Krew, "krew", || generic::run_krew_upgrade(run_type))?;
     runner.execute(Step::Gem, "gem", || generic::run_gem(&base_dirs, run_type))?;
     runner.execute(Step::Sheldon, "sheldon", || generic::run_sheldon(&ctx))?;
+    runner.execute(Step::Rtcl, "rtcl", || generic::run_rtcl(&ctx))?;
```
r-darwish, line 283: Could you please tell me which tool is this about? I could not find it.

r-darwish (changes requested):

src/main.rs
niheaven: `rtcl` is just **R tools for the command line**, and `rupdate` is one of its command, located in packages' `bin` or `win` subdir.

> `rupdate` to update all CRAN packages on your system, install missing CRAN packages listed in your collection `~/.rtcl/packages` and update all packages with a git source.

`jetpack` only update packages that installed from CRAN, `rtcl::rupdate` will update packages installed from CRAN and git resources (like GitHub, etc.).

The package is not on CRAN yet, and need to be installed from https://github.com/rdatsci/rtcl.

src/steps/generic.rs
```diff
@@ -121,6 +121,14 @@ pub fn run_jetpack(run_type: RunType) -> Result<()> {
     run_type.execute(&jetpack).args(&["global", "update"]).check_run()
 }
 
+pub fn run_rtcl(ctx: &ExecutionContext) -> Result<()> {
+    let rupdate = utils::require("rupdate")?;
+
+    print_separator("rtcl::rupdate");
```
r-darwish, line 127: Please put a friendlier name

niheaven: Yes, that what's I'm not sure... Do you think in this case `rupdate` or just `rtcl` be a good name?

r-darwish: I think rtctl is a better choice

niheaven: Using `rtcl` as separator.

r-darwish:
Thanks!

PR #550, opened 2020-10-29
Add option to use an alternative program to yay

### LINKED ISSUE #548: Paru AUR helper support
https://github.com/morganamilo/paru

See also:
https://www.reddit.com/r/archlinux/comments/jjn1c1/paru_v100_and_stepping_away_from_yay/

Arch Linux
Same as Yay
Same as Yay (so far)

silentnoodlemaster:
This PR would allow users to use something like [paru](https://github.com/morganamilo/paru) instead of yay, since they are mostly compatible with the arguments they use

would fix #548   

## Standards checklist:

- [x] The code compiles
- [x] The code passes rustfmt
- [x] The code passes clippy
- [x] The code passes tests

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -63,6 +63,8 @@
 #greedy_cask = true
 
 #[linux]
+# The name yay equivalent program, for example paru
+#yay_binary_name = "yay"
 # Arguments to pass yay when updating packages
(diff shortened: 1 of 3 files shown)


r-darwish:
Thank you for your work, but I will add support for paru without the user having to configure it.

PR #561, opened 2020-11-15
Add support for upgrading gcloud

dpetersen:
gcloud is the command line tool for interacting with Google Cloud Platform. It's the CLI you can use to do typical tasks like creating new virtual servers. But you can also use it to install "components", like installing the `kubectl` command to interact with a Google Kubernetes Engine instance, or a local emulator for some of their closed-source products. So `gcloud components update` will check for updates on gcloud itself and any components you have installed. The `--quiet` flag skips confirmation, but it still outputs a changelog for the things that were updated.

I have tested that this runs the command like I expect and outputs properly, and that Topgrade understands that it was successful. Also, I've verified that if there are no updates available, it can determine that quickly and has sane output:

```
―― 12:29:50 - gcloud ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――

All components are up to date.
```

And that it does not try to upgrade when `gcloud` is not in your `PATH`.

Thanks so much for this awesome project! I've been using it for quite a while and it was incredibly easy to add this support!

## Standards checklist:

- [x] The code compiles
- [x] The code passes rustfmt
- [x] The code passes clippy
- [x] The code passes tests

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -111,6 +111,7 @@ pub enum Step {
     Choosenim,
     Rtcl,
     Deno,
+    Gcloud,
 }
 
 #[derive(Deserialize, Default, Debug)]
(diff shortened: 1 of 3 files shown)


r-darwish (approved):

r-darwish:
Thank you

PR #585, opened 2020-12-09
Arm Brew support

r-darwish:


r-darwish:
@Byron can you test this change? You should see two brew steps running, one for ARM and one for Intel.

Byron:
Thanks for the quick turnaround! Here is the results, as run with an aarch64 topgrade binary from a aarch64 tmux in an aarch64 terminal (Alacritty):

```
―― 13:03:31 - Intel Brew ―――――――――――――――――――――――――――――――――――――――――――――――――――――――
Updating Homebrew...
Error: Cannot install in Homebrew on ARM processor in Intel default prefix (/usr/local)!
Please create a new installation in /opt/homebrew using one of the
"Alternative Installs" from:
  https://docs.brew.sh/Installation
You can migrate your previously installed formula list with:
  brew bundle dump

―― 13:03:36 - Intel Brew Cleanup ―――――――――――――――――――――――――――――――――――――――――――――――
Warning: Skipping nghttp2: most recent version 1.42.0_1 not installed
Error: Permission denied @ apply2files - /usr/local/lib/docker/cli-plugins

―― 13:03:38 - Summary ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
Brew (ARM): OK
Brew (Intel): FAILED
Git repositories: OK
oh-my-zsh: OK
rustup: OK
cargo: OK
vim: OK
npm: FAILED
Intel Brew: FAILED
Intel Brew Cleanup: FAILED
```

What's happening here is that due to brew being started from an aarch64 process, every child process is also aarch64. In order to force intel to be used, one can do something like I do in my `.zshrc`.

```zsh
alias ibrew="arch -x86_64 /usr/local/bin/brew"
```

Here is how this affects `uname` for example.

```sh
➜  topgrade git:(m1-brew) uname -m
arm64
➜  topgrade git:(m1-brew) arch -x86_64 uname -m
x86_64
```

Inversely, I don't know of a way to switch from an INTEL shell to ARM, so an INTEL topgrade is likely to fail running the brew ARM version instead. Probably it's worth checking for that to avoid running ARM versions on intel.

r-darwish:
Did the arm version of brew ran as expected?

Byron:
> Did the arm version of brew ran as expected?

Yes

> Does topgrade need to do the opposite as well? (Run the ARM brew with arch if Topgrade is compiled with x86_64)

Yes, I believe it would have to. With `arch` (see the manual included in the details section) ~~that doesn't seem to be possible~~. You can upgrade to ARM using `arch -arm64e <program>` - the flag seems to be undocumented, but works.

The screenshot shows nesting arm, intel and arm.
Screenshot 2020-12-09 at 16 53 41

However, this only works if the target program actually is a universal program, otherwise you see `arch -arm64e /usr/local/bin/bash` _(intel bash installed with brew)_ : `arch: posix_spawnp: /usr/local/bin/bash: Bad CPU type in executable`.

My recommendation is to keep this as special handling for brew for now, and leave everything else to the architecture of the child program, unless there are other package managers with well known ARM behaviour and side by side installations.

Whether or not to try anything special could be gated behind the outcome of `uname -m` so special handling is only applied if topgrade learns that ARM is actually available.

> Is this something specific to Brew or will other stuff start breaking if running from the ARM version of Topgrade unless ran with arch?

I don't have experience with other package managers, but from what I can tell, INTEL compiled topgrade child processes would not cause trouble if these are in the 'right' architecture. Thus I would say it's specific to `brew` as it's a ruby program with the system ruby being a universal binary including the intel and the ARM version. Which one gets chosen depends on the parent process. Most other tooling, unless provided by Apple, will likely be either INTEL or ARM, so it's hard to tell what's 'right' in the moment.

**Phew**, sorry for the somewhat messy write-up, that sounds like a lot of complexity but maybe you can untangle it nonetheless.
I am happy to test the ARM version whenever you give me a shout.

<details>

```
ARCH(1)                   BSD General Commands Manual                  ARCH(1)

NAME
     arch -- print architecture type or run selected architecture of a universal binary

SYNOPSIS
     arch
     arch [-32] [-64] [[-arch_name | -arch arch_name]...] [-c] [-d envname]... [-e envname=value]... [-h] prog [args ...]

DESCRIPTION
     The arch command with no arguments, displays the machine's architecture type.

     The other use of the arch command is to run a selected architecture of a universal binary.  A universal binary contains code that can run on different architectures.  By
     default, the operating system will select the architecture that most closely matches the processor type.  A 64-bit architecture is preferred over a 32-bit architecture on a
     64-bit processor, while only 32-bit architectures can run on a 32-bit processor.

     When the most natural architecture is unavailable, the operating system will try to pick another architecture.  On 64-bit processors, a 32-bit architecture is tried.  Other-
     wise, no architecture is run, and an error results.

     The arch command can be used to alter the operating system's normal selection order.  The most common use is to select the 32-bit architecture on a 64-bit processor, even if
     a 64-bit architecture is available.

     The arch_name argument must be one of the currently supported architectures:

           i386     32-bit intel

           x86_64   64-bit intel

           x86_64h  64-bit intel (haswell)

     Either prefix the architecture with a hyphen, or (for compatibility with other commands), use -arch followed by the architecture.

     If more than one architecture is specified, the operating system will try each one in order, skipping an architecture that is not supported on the current processor, or is
     unavailable in the universal binary.

     The other options are:

     -32               Add the native 32-bit architecture to the list of architectures.

     -64               Add the native 64-bit architecture to the list of architectures.

     -c                Clears the environment that will be passed to the command to be run.

     -d envname        Deletes the named environment variable from the environment that will be passed to the command to be run.

     -e envname=value  Assigns the given value to the named environment variable in the environment that will be passed to the command to be run.  Any existing environment vari-
                       able with the same name will be replaced.

     -h                Prints a usage message and exits.

     The prog argument is the command to run, followed by any arguments to pass to the command.  It can be a full or partial path, while a lone name will be looked up in the
     user's command search path.

     If no architectures are specified on the command line, the arch command takes the basename of the prog argument and searches for the first property list file with that base-
     name and the .plist suffix, in the archSettings sub-directory in each of the standard domains, in the following order:

           ~/Library/archSettings         User settings

           /Library/archSettings          Local settings

           /Network/Library/archSettings  Network settings

           /System/Library/archSettings   System settings

     This property list contains the architecture order preferences, as well as the full path to the real executable.  For examples of the property list format, look at the files
     in /System/Library/archSettings.

   Example
     On an intel processor:

           % perl -MConfig -e 'printf "%s\n", $Config{byteorder}'
           1234

     shows the intel little endian byte order.

   Making links to the arch command
     When a link is made to arch command with a different name, that name is used to find the corresponding property list file.  Thus, other commands can be wrapped so that they
     have custom architecture selection order.

     Because of some internal logic in the code, hard links to the arch command may not work quite right.  It is best to avoid using hard links, and only use symbolic links to the
     arch command.

   Environment
     The environment variable ARCHPREFERENCE can be used to provide architecture order preferences.  It is checked before looking for the corresponding property list file.

     The value of the environment variable ARCHPREFERENCE is composed of one or more specifiers, separated by semicolons.  A specifier is made up of one, two or three fields, sep-
     arated by colons.  Architectures specified in order, are separated by commas and make up the last (mandatory) field.  The first field, if specified, is a name of a program,
     which selects this specifier if that name matches the program name in question.  If the name field is empty or there is no name field, the specifier matches any program name.
     Thus, ordering of specifiers is important, and the one with no name should be last.

     When the arch command is called directly, the prog name provides the path information to the executable (possibly via the command search path).  When a name is specified in a
     ARCHPREFERENCE specifier, the path information can alternately be specified as a second field following the name.  When the arch command is called indirectly via a link, this
     path information must be specified.  If not specified as a second field in a specifier, the executable path will be looked up in the corresponding property list file.

   Example ARCHPREFERENCE Values
     i386,x86_64,x86_64h
           A specifier that matches any name.

     foo:i386,x86_64,x86_64h
           A specifier that matches the program named foo (the full executable path is in the foo.plist file).

     foo:/op/bin/boo:i386,x86_64,x86_64h
           A specifier with all fields specified.

     baz:i386;x86_64;x86_64h
           A specifier for baz and a second specifier that would match any other name.

BUGS
     Running the arch command on an interpreter script may not work if the interpreter is a link to the arch command, especially if a 64-bit architecture is specified (since the
     arch command is 2-way universal, 32-bit only).

SEE ALSO
     machine(1)

Mac OS X                         July 8, 2010                         Mac OS X
```

r-darwish:
Thanks for the detailed answer. I'll submit a patch to do as you suggested.
@claui I'll by happy to hear your opinion about this as well.

r-darwish:
@Byron I pushed some changes. Can you please check?
If I got it right then it should run both brews without any issues whether Topgrade itself is ARM or x86. Could you please test it in both targets?

Byron:
### ARM test result

~~Using `cargo run` from an ARM shell, it looks like brew intel didn't work as expected for some reason.~~ tl;dr: it works! The whole journey is in the details. As I kicked INTEL rustup and painfully switched to an ARM toolchain, I lost my ability to compile intel binaries with Rust. Maybe intel can be tested by someone who has an intel mac?
What I could verify is that running the ARM binary from an INTEL shell works as expected as well.


<details>

```
―― 08:29:32 - Vim ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
Plugins upgraded

―― 08:29:32 - Node Package Manager ―――――――――――――――――――――――――――――――――――――――――――――
env: node: No such file or directory

―― 08:29:32 - Intel Brew ―――――――――――――――――――――――――――――――――――――――――――――――――――――――
Updating Homebrew...
Error: Cannot install in Homebrew on ARM processor in Intel default prefix (/usr/local)!
Please create a new installation in /opt/homebrew using one of the
"Alternative Installs" from:
  https://docs.brew.sh/Installation
You can migrate your previously installed formula list with:
  brew bundle dump

―― 08:29:46 - Intel Brew Cleanup ―――――――――――――――――――――――――――――――――――――――――――――――
Error: Permission denied @ apply2files - /usr/local/lib/docker/cli-plugins

―― 08:29:48 - Summary ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
Brew (ARM): OK
Brew (Intel): FAILED
Git repositories: OK
oh-my-zsh: OK
rustup: OK
cargo: OK
vim: OK
npm: FAILED
Intel Brew: FAILED
Intel Brew Cleanup: FAILED
➜  topgrade git:(m1-brew) file ./target/debug/topgrade
./target/debug/topgrade: Mach-O 64-bit executable arm64
```

Running it manually did not fail though, but there was nothing to upgrade either.

```
➜  topgrade git:(m1-brew) arch -x86_64 /usr/local/bin/brew upgrade
Updating Homebrew...
```

When running it plainly I could reproduce the error though…

```
➜  topgrade git:(m1-brew) /usr/local/bin/brew upgrade
Updating Homebrew...
Error: Cannot install in Homebrew on ARM processor in Intel default prefix (/usr/local)!
Please create a new installation in /opt/homebrew using one of the
"Alternative Installs" from:
  https://docs.brew.sh/Installation
You can migrate your previously installed formula list with:
  brew bundle dump
```

…which might hint at the `arch` invocation performed by topgrade not doing what it should when launched in topgrade's execution context?

What about this program?

```rust
use std::process::Command;

fn main() {
    Command::new("arch")
        .arg("-x86_64")
        .arg("/usr/local/bin/brew")
        .arg("upgrade")
        .spawn()
        .unwrap()
        .wait()
        .unwrap();
}
```

It seems to work as expected.
```
➜  topgrade git:(m1-brew) ✗ cargo run
   Compiling topgrade v6.0.2 (/Users/byron/dev/topgrade)
    Finished dev [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/topgrade`
Updating Homebrew...
==> Auto-updated Homebrew!
Updated 1 tap (homebrew/core).
==> Updated Formulae
Updated 1 formula.
```

Maybe there is something unexpected happening in the topgrade invocation of brew, so running `cargo run -- -v` reveals the following:

```
―― 09:03:55 - Intel Brew ―――――――――――――――――――――――――――――――――――――――――――――――――――――――
 2020-12-11T01:03:55.540Z DEBUG topgrade::executor   > Running "/bin/zsh" "-c" "/usr/local/bin/brew upgrade"
Updating Homebrew...
Error: Cannot install in Homebrew on ARM processor in Intel default prefix (/usr/local)!
Please create a new installation in /opt/homebrew using one of the
"Alternative Installs" from:
  https://docs.brew.sh/Installation
You can migrate your previously installed formula list with:
  brew bundle dump
 2020-12-11T01:04:04.865Z DEBUG topgrade::runner     > Step "Intel Brew Cleanup"

―― 09:04:04 - Intel Brew Cleanup ―――――――――――――――――――――――――――――――――――――――――――――――
 2020-12-11T01:04:04.865Z DEBUG topgrade::executor   > Running "/bin/zsh" "-c" "/usr/local/bin/brew cleanup"
Error: Permission denied @ apply2files - /usr/local/lib/docker/cli-plugins
```

The arch invocation is notably missing, however, further up the log there is the invocation as we would expect it (using arch), and that made me realize I still had my own custom commands in the configuration. After removing it, everything looked like it worked properly.

Byron:
Correction: Launching  ARM `topgrade` from an INTEL shell fails to run ARM brew with the following error:

```
―― 09:40:02 - Brew (ARM) ―――――――――――――――――――――――――――――――――――――――――――――――――――――――
 2020-12-11T01:40:03.926Z DEBUG topgrade::executor > Running "/opt/homebrew/bin/brew" "update"
Updated 1 tap (homebrew/cask).
==> New Casks
pictogram
==> Updated Casks
hstracker                messenger                microsoft-edge           private-internet-access
 2020-12-11T01:41:36.966Z DEBUG topgrade::executor > Running "/opt/homebrew/bin/brew" "upgrade" "--ignore-pinned"
Updating Homebrew...
Error: Cannot install in Homebrew on Intel processor in ARM default prefix (/opt/homebrew)!
```

This seems to be due to it not using `arch` (for good reason), yet it appears the child process inherits the INTEL architecture of the parent shell instead of the ARM architecture of the parent process.
This seems to mean that the [session leader's](https://www.win.tue.nl/~aeb/linux/lk/lk-10.html) architecture is the determining factor, in that case the INTEL shell. This behaviour probably makes sense if a GUI app is the session leader as well.

I think this can be countered by always running brew through `arch` if topgrade is an ARM binary.

r-darwish:
https://github.com/r-darwish/topgrade/blob/8054374bc2d1ea40445cc7c3b44a90269c807d49/src/steps/os/unix.rs#L54-L58

Actually it was supposed to use arch, but I'm not sure why it didn't. Maybe `target_arch` is something other than `x86_64`?

Byron:
Sorry for the late reply.

> Actually it was supposed to use arch, but I'm not sure why it didn't. Maybe target_arch is something other than x86_64?

I believe both arch names are correct, one of them is also mentioned in [the Rust reference](https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg_attr-attribute), the other one shows up in the debug output of topgrade itself:

```
 2020-12-15T06:09:07.323Z DEBUG topgrade         > Version: 6.0.2
 2020-12-15T06:09:07.323Z DEBUG topgrade         > OS: aarch64-apple-darwin
```

As my machine is moving target itself, I managed to remove the intel brew entirely and now so can't really test everything right now. However, running it in verbose mode now yields:

```
➜  topgrade git:(m1-brew) cargo run -- -v
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
     Running `target/debug/topgrade -v`
 2020-12-15T06:14:57.155Z DEBUG topgrade::config > Configuration at /Users/byron/.config/topgrade.toml
 2020-12-15T06:14:57.158Z DEBUG topgrade::config > Loaded configuration: ConfigFile { pre_commands: None, post_commands: None, commands: None, git_repos: None, predefined_git_repos: None, disable: Some([System]), ignore_failures: None, remote_topgrades: None, remote_topgrade_path: None, ssh_arguments: None, git_arguments: None, tmux_arguments: None, set_title: None, assume_yes: None, yay_arguments: None, no_retry: Some(true), run_in_tmux: None, cleanup: Some(true), notify_each_step: None, accept_all_windows_updates: None, only: None, composer: None, brew: None, linux: None, git: None, windows: None, vagrant: None }
 2020-12-15T06:14:57.158Z DEBUG topgrade         > Version: 6.0.2
 2020-12-15T06:14:57.158Z DEBUG topgrade         > OS: aarch64-apple-darwin
 2020-12-15T06:14:57.158Z DEBUG topgrade         > Args { inner: ["target/debug/topgrade", "-v"] }
 2020-12-15T06:14:57.159Z DEBUG topgrade         > Binary path: Ok("/Users/byron/dev/topgrade/target/debug/topgrade")
 2020-12-15T06:14:57.159Z DEBUG topgrade         > Self Update: false
 2020-12-15T06:14:57.159Z DEBUG topgrade::utils  > Detected "/usr/bin/git" as "git"
 2020-12-15T06:14:57.159Z DEBUG topgrade::utils  > Cannot find "doas"
 2020-12-15T06:14:57.159Z DEBUG topgrade::utils  > Detected "/usr/bin/sudo" as "sudo"
 2020-12-15T06:14:57.159Z DEBUG topgrade::utils  > Cannot find "pwsh"
 2020-12-15T06:14:57.159Z DEBUG topgrade::utils  > Cannot find "powershell"
 2020-12-15T06:14:57.159Z DEBUG topgrade::runner > Step "Brew"
 2020-12-15T06:14:57.159Z DEBUG topgrade::utils  > Detected "/opt/homebrew/bin/brew" as "/opt/homebrew/bin/brew"

―― 14:14:57 - Brew ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
 2020-12-15T06:14:57.611Z DEBUG topgrade::executor > Running "/opt/homebrew/bin/brew" "update"
Already up-to-date.
 2020-12-15T06:16:03.165Z DEBUG topgrade::executor > Running "/opt/homebrew/bin/brew" "upgrade" "--ignore-pinned"
Updating Homebrew...
^CWarning: You are running macOS on a arm64 CPU architecture.
We do not provide support for this (yet).
Reinstall Homebrew under Rosetta 2 until we support it.
You will encounter build failures with some formulae.
Please create pull requests instead of asking for help on Homebrew's GitHub,
Twitter or any other official channels. You are responsible for resolving
any issues you experience while you are running this
unsupported configuration.

 2020-12-15T06:23:01.094Z DEBUG topgrade::executor > Running "/opt/homebrew/bin/brew" "cleanup"
 2020-12-15T06:23:03.742Z DEBUG topgrade::runner   > Step "Brew"
thread 'main' panicked at 'Brew already reported', src/report.rs:37:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Could it be that there is accidental breakage in case there is only one homebrew installation, or is there some mess on my system that's causing this 😅?

src/steps/os/unix.rs
```diff
@@ -14,6 +14,53 @@ use std::path::PathBuf;
...
+        match self {
+            BrewVariant::MacArm if both_exists => "Brew (ARM)",
+            BrewVariant::MacIntel if both_exists => "Brew (Intel)",
+            _ => "Brew",
+        }
+    }
+
+    fn execute(self, run_type: RunType) -> Executor {
+        match self {
+            BrewVariant::MacIntel if cfg!(target_arch = "aarch64") => {
```
Byron, line 49: I can definitely verify that this target arch specifier is correct.

r-darwish:
If only one brew is present then logic should be the same except that it should just appear as "Brew" instead of "Brew (ARM)" or "Brew (Intel)".
What I see in your latest post is that you run Topgrade as aarch and it runs the ARM version of brew without using "arch". Isn't this the expected thing to do?

Byron:
That's correct and expected. A few lines below though there is this panic:

```
 2020-12-15T06:23:03.742Z DEBUG topgrade::runner   > Step "Brew"
thread 'main' panicked at 'Brew already reported', src/report.rs:37:13
```

r-darwish:
Panic should be fixed now

Byron:
It is, thanks a lot!

From my point of view, it does work as expected on an M1 mac with only Homebrew for ARM installed, and from reading the code it should do the right thing with Arch when running the intel homebrew on an M1 Mac as well.

claui:
> @claui I'll by happy to hear your opinion about this as well.

I’ve been unable to test this due to workload. I’ve inspected the changes and it all looks fine.

r-darwish:
Great. I merged it. I suppose we'll get more feedback as more people will get an ARM Mac

PR #623, opened 2021-02-09
Surpress explicit news printing when paru is used

miterion:
This is just a small change which supresses the fetching of arch news when paru is used, since it is done automatically when  `paru` is executed. I am not that well versed in rust, hope it is still an okayish way to do this.

## Standards checklist:

- [x] The code compiles
- [x] The code passes rustfmt
- [x] The code passes clippy
- [x] The code passes tests

### DIFF
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -157,13 +157,14 @@ fn upgrade_arch_linux(ctx: &ExecutionContext) -> Result<()> {
     debug!("Running Arch update with path: {:?}", path);
 
     if let Some(yay) = which("yay").or_else(|| which("paru")) {
-        run_type
-            .execute(&yay)
-            .arg("-Pw")
-            .spawn()
-            .and_then(|mut p| p.wait())
-            .ok();
-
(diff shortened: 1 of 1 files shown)


src/steps/os/linux.rs
```diff
@@ -157,13 +157,14 @@ fn upgrade_arch_linux(ctx: &ExecutionContext) -> Result<()> {
     debug!("Running Arch update with path: {:?}", path);
 
     if let Some(yay) = which("yay").or_else(|| which("paru")) {
-        run_type
```
r-darwish, line 160: I rather have some thing like 
```rust 
let yay = which("yay")
if let Some(yay) = yay {
// Show news
}

if let Some(yay) = yay.or_else(|| which ("paru")) {
// Upgrade
}
````

r-darwish (changes requested):
Thank you. Please see my comment

src/steps/os/linux.rs
miterion: Thanks for the feedback, I implemented it in 4b11094

PR #628, opened 2021-02-12
added support for Visual Studio Code (VSC) extension update

AlexanderWillner:
## Standards checklist:

- [x] The code compiles
- [x] The code passes rustfmt
- [x] The code passes clippy
- [x] The code passes tests (no tests implemented)

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -86,15 +86,20 @@ pub fn run_sheldon(ctx: &ExecutionContext) -> Result<()> {
     target_os = "dragonfly"
 )))]
 pub fn run_vscode(run_type: RunType) -> Result<()> {
-    let _vscode = utils::require("code")?;
+    let vscode = utils::require("code")?;
 
     print_separator("Visual Studio Code");
 
-    run_type
-        .execute(shell())
-        .arg("-c")
-        .arg("for ext in $(code --list-extensions); do code --force --install-extension \"$ext\"; done")
-        .check_run()
+    let plugins = run_type.execute(&vscode).args(&["--list-extensions"]).check_output()?;
+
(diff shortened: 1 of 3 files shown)


SuperSandro2000:
Can we rename this to VSCode? VSC is not as common and probably confuses a few people what it is.

src/steps/generic.rs
```diff
@@ -85,6 +85,18 @@ pub fn run_sheldon(ctx: &ExecutionContext) -> Result<()> {
...
 )))]
+pub fn run_vscode(run_type: RunType) -> Result<()> {
+    let _vscode = utils::require("code")?;
+
+    print_separator("Visual Studio Code");
+
+    run_type
+        .execute(shell())
+        .arg("-c")
+        .arg("for ext in $(code --list-extensions); do code --force --install-extension \"$ext\"; done")
```
r-darwish, line 96: Thanks, but we need to avoid using a Unix shell because it won't work in Windows

r-darwish (changes requested):

r-darwish:
Will this download only new versions of packages or will it cause vscode to redownload all of your extensions every time Topgrade is invoked?

SuperSandro2000:
> Will this download only new versions of packages or will it cause vscode to redownload all of your extensions every time Topgrade is invoked?

it will query the vscode marketplace for the latest version and if that is already installed nothing happens. This will take some time depended on the amount of extensions you have installed. For 43 extensions this took me 60 seconds.

src/steps/generic.rs
AlexanderWillner: What's the best practice here? Implementing the needed loop in rust I assume?

Issue #632, opened 2021-02-14
topgrade panics with the --dry-run flag when VS Code extension updates are enabled


SanchithHegde:
**What did you expect to happen?**
`topgrade` shouldn't panic when the `--dry-run` flag is passed and VS Code extension updates are enabled.

**What actually happened?**
`topgrade` panics when the `--dry-run` flag is passed and VS Code extension updates are enabled. I disabled VS Code extension updates and I no longer face the issue. Normal updates (just running `topgrade` without any flags) are unaffected by this issue.

**Additional details:**
- Which operating system or Linux distribution are you using? Arch Linux
- How did you install Topgrade? AUR package, `topgrade 6.5.1-1`

<details>
<summary>Output of <code>RUST_BACKTRACE=1 cargo r -- --dry-run --verbose --only vs_code</code></summary>

(Running with `cargo` for debug output.)

<pre>
❯ RUST_BACKTRACE=1 cargo r -- --dry-run --verbose --only vs_code            
   Compiling topgrade v6.5.1 (/home/username/Downloads/topgrade)
    Finished dev [unoptimized + debuginfo] target(s) in 1.61s
     Running `target/debug/topgrade --dry-run --verbose --only vs_code`
 2021-02-14T20:18:45.924Z DEBUG topgrade::config > Configuration at /home/username/.config/topgrade.toml
 2021-02-14T20:18:45.924Z DEBUG topgrade::config > Loaded configuration: ConfigFile { pre_commands: None, post_commands: None, commands: None, git_repos: None, predefined_git_repos: None, disable: Some([Asdf, Gem]), ignore_failures: None, remote_topgrades: None, remote_topgrade_path: None, ssh_arguments: None, git_arguments: None, tmux_arguments: None, set_title: None, assume_yes: None, yay_arguments: None, no_retry: None, run_in_tmux: None, cleanup: None, notify_each_step: None, accept_all_windows_updates: None, bashit_branch: None, only: None, composer: None, brew: None, linux: None, git: None, windows: None, vagrant: None }
 2021-02-14T20:18:45.925Z DEBUG topgrade::utils  > Detected "/usr/bin/notify-send" as "notify-send"
 2021-02-14T20:18:45.925Z DEBUG topgrade         > Version: 6.5.1
 2021-02-14T20:18:45.925Z DEBUG topgrade         > OS: x86_64-unknown-linux-gnu
[...]

PR #648, opened 2021-02-19
Fix panic when dry running micro

SanchithHegde:
Dry running `micro` would produce an error message similar to that in #632.

As of now, the `run_type` parameter is not used anywhere in the function and I've prefixed it with an underscore to keep the interface similar to other functions in the module. In case the parameter can be removed, let me know and I'll update it, or you can go ahead and make the necessary changes.

## Standards checklist:

- [x] The code compiles
- [x] The code passes rustfmt
- [x] The code passes clippy
- [x] The code passes tests

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -110,12 +110,15 @@ pub fn run_vscode(run_type: RunType) -> Result<()> {
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


r-darwish:
Sorry for closing this one, but with the solution dry run won't be dry with micro. I'm closing this in favor of #653 as we need to find some more generic solution to this problem.

PR #650, opened 2021-02-21
Fix nix on multi-user setup (fix #646)

### LINKED ISSUE #646: Failed to update multi-user mode nix

`topgrade` succeeds to update multi-user mode nix

**What actually happened?**

`topgrade` fails as follows.

```
―― 15:21:37 - Nix ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
sudo: nix: command not found
```

The `nix` command is installed under `/nix` by default, so the `PATH` environment variable must be set for it to run.
In multi-user mode nix, the `PATH` is set by `/etc/profile.d/nix.sh`, but the `sudo` command will not load this script by default, so `nix` will fail to run.
To avoid this problem, you need to execute the `nix` command with the profile script loaded, like `sudo -i nix`.

**Additional details:**
  - CentOS7
  -  `nix-env -i topgrade` or `cargo install topgrade`

Result of `nix --dry-run`

```
―― 15:24:45 - Nix ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
Dry running: /usr/bin/sudo nix upgrade-nix
Dry running: /nix/var/nix/profiles/default/bin/nix-channel --update
Dry running: /nix/var/nix/profiles/default/bin/nix-env --upgrade
```

r-darwish:
## Standards checklist:

- [ ] The code compiles
- [ ] The code passes rustfmt
- [ ] The code passes clippy
- [ ] The code passes tests

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -175,7 +175,10 @@ pub fn run_nix(ctx: &ExecutionContext) -> Result<()> {
 
     if multi_user {
         if let Some(sudo) = ctx.sudo() {
-            run_type.execute(&sudo).arg("nix").arg("upgrade-nix").check_run()?;
+            run_type
+                .execute(&sudo)
+                .args(&["-i", "nix", "upgrade-nix"])
+                .check_run()?;
         } else {
             print_warning("Need sudo to upgrade Nix");
         }


r-darwish:
@gifnksm can you please test this change?

gifnksm:
@r-darwish Thank you for the fix. It works fine in my environment.

However, I am afraid that it may not work in the environment where the `doas` command is installed.
The value of the local variable `sudo` comes from the return value of [`crate::utils::sudo`]. This function will return `doas` instead of `sudo` in environments where the `doas` command is installed.

According to [man], `doas` does not have a `-i` option, so this fix may not work well in such environments.

[`crate::utils::sudo`]: https://github.com/r-darwish/topgrade/blob/f73c3ef31a766e6cc71c05654c12714b5c3eabed/src/utils.rs#L105
[man]: https://man.archlinux.org/man/doas.1.en

r-darwish:
I think we can live with that at the moment. If someone complains we will look for a solution.

PR #660, opened 2021-02-24
Drop the Go step

### LINKED ISSUE #659: steps/go: topgrade breaks updating the Go packages (or whatever) with Go 1.16
Updating the Go packages correctly or whatever should be updated here.

**What actually happened?**
Failed to update the Go packages or whatever should be updated.
Message: `go get: cannot match "all": working directory is not part of a module`

**Additional details:**
  - `Arch linux running 5.10.16 linux kernel`
  - `yay -S topgrade` (AUR https://aur.archlinux.org/packages/topgrade/)

**Actual description:**
What is the intention of the Go step?

As of `topgrade --dry-run` topgrade tries to run `/usr/bin/go get -u all in /home/link/go` on the Go step (also this was what I saw in the code base). As of Go 1.16 the behavior of `go get` changed. In the previous versions `go get` could be used to eather download a Go package or to install a Go tool (if the package could be built as a command). Now with Go 1.16 `go get` does not support the installation of commands anymore but can only download the requested Go packages (libraries). To install a Go command/tool now you have to use `go install` explicitly. That's why `go get -u all` does not work anymore if the `go` command is executed outside of a Go module project.

Currently this behavior can be changed temporarily by using `GO111MODULE=auto|off go get -u all` but as of https://blog.golang.org/go116-module-changes Go 1.17 will ignore the GO111MODULE completly and so the GOPATH mode also. In other words Go 1.17 will drop the GOPATH mode completly and the GOMODULE mode will not only be the default but the only option to manage and develop Go projects.

So now I wonder what the initial intention was behind the Go step of topgrade so that one might be able to tackle this issue somehow, eather by changing the behaivor according to some kind of wanted or wishable behaivor or even by dropping the step completly as it would not have any effect in the future.

Akeshihiro:
With the release of Go 1.16 the behavior of `go get` has been changed.
In previous Go versions `go get` was used not only to add module
dependencies but also to install Go tools.
As of Go 1.16 `go get` can only add and upgrade module dependencies.
To install Go tools now the `go install` command has to be used.

Further on Go 1.16 enabled the GOMODULE mode by default and will drop
the GOPATH mode completly in Go 1.17.
So the package definition `all` like in `go get -u all` does not work
anymore if the PWD is outside of a Go module project.
Because of this `go list all` also does not work for the same reason.
That being said it seems that currently there is no way to get a list of
all installed Go tools or packages at the GOPATH level.

So the only possible solution to determine the installed Go tools and
also to update them would be by inspecting the `go env GOBIN` directory
as well as the `go env GOMODCACHE` sub-directories and to filter the
results according to their possible name-to-package boundaries.
As this approach seems to be very ugly and also not to be very safe or
stable and Go currently does not support any kind of automated upgrades
of installed Go tools it is best to drop the Go step for now until Go
implements some kind of Go tool upgrade feature.

Fixes #659

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -82,7 +82,6 @@ pub enum Step {
     Gcloud,
     Gem,
     GitRepos,
-    Go,
     HomeManager,
     Jetpack,
     Krew,
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -101,6 +101,7 @@ impl Executor {
         self
     }
 
+    #[allow(dead_code)]
     /// See `std::process::Command::remove_env`
     pub fn env_remove<K>(&mut self, key: K) -> &mut Executor
     where
--- a/src/main.rs
+++ b/src/main.rs
@@ -286,7 +286,6 @@ fn run() -> Result<()> {
     runner.execute(Step::Choosenim, "choosenim", || generic::run_choosenim(&ctx))?;
     runner.execute(Step::Cargo, "cargo", || generic::run_cargo_update(run_type))?;
     runner.execute(Step::Flutter, "Flutter", || generic::run_flutter_upgrade(run_type))?;
-    runner.execute(Step::Go, "Go", || generic::run_go(run_type))?;
     runner.execute(Step::Emacs, "Emacs", || emacs.upgrade(run_type))?;
     runner.execute(Step::Opam, "opam", || generic::run_opam_update(run_type))?;
     runner.execute(Step::Vcpkg, "vcpkg", || generic::run_vcpkg_update(run_type))?;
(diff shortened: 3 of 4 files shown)


r-darwish (changes requested):
Please also remove `Step::Go` from `config.rs`

r-darwish:
Good riddance 😄 
Thank you

Akeshihiro:
Thank you for this great tool. I'm using topgrade for some month now across multiple devices and OSes and it works like a charm. Sometimes making some upgrades really is a nightmare! So yeah, very happy with topgrade :-)

thomwiggers:
Go has not been removed from the wiki pages describing supported steps yet.

r-darwish:
@thomwiggers fixed

PR #663, opened 2021-03-01
Support for PIP3 (self update)

AlexanderWillner:
## Standards checklist:

- [x] The code compiles (`cargo b`)

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -226,7 +226,7 @@ pub fn run_pip3_update(run_type: RunType) -> Result<()> {
 
     run_type
         .execute(&pip3)
-        .args(&["install", "--upgrade", "pip"])
+        .args(&["install", "--upgrade", "--user", "pip"])
         .check_run()
 }
 
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -220,6 +220,13 @@ pub fn run_pipx_update(run_type: RunType) -> Result<()> {
     run_type.execute(&pipx).arg("upgrade-all").check_run()
 }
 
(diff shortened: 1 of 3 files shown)


src/steps/generic.rs
```diff
@@ -220,6 +220,16 @@ pub fn run_pipx_update(run_type: RunType) -> Result<()> {
     run_type.execute(&pipx).arg("upgrade-all").check_run()
 }
 
+pub fn run_pip3_update(run_type: RunType) -> Result<()> {
+    let pip3 = utils::require("pip3")?;
+    print_separator("pip3");
+
+    run_type
+        .execute(&pip3)
+        .args(&["install", "--upgrade", "pip"])
```
SuperSandro2000, line 229: This should use --user to not overwrite the system pip.

AlexanderWillner: I believe it will not overwrite the system pip, as this is usually not writable for the user (e.g. on macOS I get a `Defaulting to user installation because normal site-packages is not writeable`). If the user wants actually to upgrade the system version he/she can invoke `topgrade` with `sudo`, no?

r-darwish: This is true in Linux, however, this is what you get in Windows when not using `--user`:

```
ERROR: Could not install packages due to an EnvironmentError: [WinError 5] Access is denied: 'c:\\pyth on39\\lib\\site-packages\\pip-20.2.3.dist-info\\entry_points.txt' Consider using the `--user` option or check the permissions.
```

So we should specify `--user` explicitly

r-darwish (changes requested):

PR #683, opened 2021-04-06
Use pkcon over apt with kde neon

RoryO:
KDE Neon is Ubuntu based and uses `pkcon` and PackageKit instead of `apt`. It errors if you try and use `apt`

~~~
On KDE neon you should use `pkcon update` to install updates.
If you absolutely must use apt you do have to use dist-upgrade or full-upgrade in place of the upgrade command.
  https://neon.kde.org/faq#command-to-update
Abort.
~~~

In theory anything can use PackageKit, in practice it doesn't look like anything other than KDE Neon does by default these days. I made a note of this in the comments for future spelunkers if this ever comes up again.

I have never worked with rust before, sorry.

## Standards checklist:

### DIFF
--- a/src/steps/os/linux.rs
+++ b/src/steps/os/linux.rs
@@ -353,28 +353,43 @@ fn upgrade_gentoo(ctx: &ExecutionContext) -> Result<()> {
 
 fn upgrade_debian(ctx: &ExecutionContext) -> Result<()> {
     if let Some(sudo) = &ctx.sudo() {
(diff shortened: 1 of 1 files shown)


r-darwish (changes requested):
Thanks for the pull request. I rather that we detect KDE neon as a seperate distro that has its own function instead of changing `upgrade_debian`

RoryO:
Okay, I'll look into that.

RoryO:
Should be better now.

PR #707, opened 2021-05-04
No need to run which in WSL (fix #694)

### LINKED ISSUE #694: topgrade steps fails on wsl when launched by windows

I would like that launching this on windows
``` powershell
topgrade --only wsl
```

is equal to
``` bash
wsl
topgrade
```

**What actually happened?**
Launching topgrade via windows, will cause a failure with some packages: yarn, npm, pip3 in my case
Launching topgrade via linux shell it works like expected

**Additional details:**
- Windows 10
- WSL2, Ubuntu
- python and node installed via asdf (that requires .bashrc/.profile loading)
- default shell on wsl: bash (also tried fish with same results)

It seems that profile cannot be loaded properly.

`topgrade -v` does not shows intresting logs.

r-darwish:
## Standards checklist:

### DIFF
--- a/src/steps/os/windows.rs
+++ b/src/steps/os/windows.rs
@@ -52,15 +52,8 @@ pub fn run_scoop(cleanup: bool, run_type: RunType) -> Result<()> {
 
 pub fn run_wsl_topgrade(ctx: &ExecutionContext) -> Result<()> {
     let wsl = require("wsl")?;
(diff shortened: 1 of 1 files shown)


r-darwish:
@webartoli can you check if this change is working as expected even if you echo something in your `~/.profile`?

PR #712, opened 2021-05-10
makes rpm-ostree configurable

### LINKED ISSUE #710: existence of rpm-ostree doesn't mean 'system' can use it

DNF to be used for the `system` module (`/usr/bin/dnf-3`)

**What actually happened?**

Topgrade [detects that my system has the file `/usr/bin/rpm-ostree`](https://github.com/r-darwish/topgrade/blob/69e05d7ceaf2bf04d6d07c427247c1d67ed012b4/src/steps/os/linux.rs#L240-L248) and tries to use that to upgrade. However it fails since the system isn't an ostree based system.

**Additional details:**
- Which operating system or Linux distribution are you using?: Fedora 34
- How did you install Topgrade?: Cargo

<details><pre>
Will get the output once I get back to that system.
</pre></details>

Thanks for Topgrade, I've been using it for quite a while and really love it!

kbknapp:
This commit makes it possible to *not* to use `rpm-ostree` even on
systems where it is detected.

This commit is fully backwards compatible with previous releases, and
without changing the config file all previous behaviors are the exact
same.

This commit adds the `rpm_ostree` field in the `[linux]` table of the
configuration, and defaults to `true`. This means Topgrade will first
check if `/usr/bin/rpm-ostree` exists, and only if so then check if the
user does not want to use `rpm-ostree` via the configuration. If the
user *does not* want to use `rpm-ostree`, then normal operation
continues checking for DNF or YUM.

This makes it possible for people where `rpm-ostree` is installed, but
where the system is not an `ostree` based distribtuion. This happens
when people are using things like `osbuild-composer` to build images,
or Cockpit with the Compose feature enabled (which uses
`osbuild-composer` internally).

An alternative to this commit would be to make the config field a
negative such as `no_rpm_ostree`, however that goes against the norm in
other fields.

Closes #710

## Standards checklist:

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -70,6 +70,7 @@
 #emerge_sync_flags = "-q"
 #emerge_update_flags = "-uDNa --with-bdeps=y world"
 #redhat_distro_sync = false
+#rpm_ostree = false
 
 [windows]
 # Manually select Windows updates
--- a/src/config.rs
+++ b/src/config.rs
@@ -159,6 +159,7 @@ pub struct Linux {
     apt_arguments: Option<String>,
     enable_tlmgr: Option<bool>,
     redhat_distro_sync: Option<bool>,
+    rpm_ostree: Option<bool>,
     emerge_sync_flags: Option<String>,
     emerge_update_flags: Option<String>,
 }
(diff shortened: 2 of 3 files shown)


kbknapp:
That's a potential, but I believe that having a mechanism for both detecting those distros (as each my have different ways of getting that information) and maintaining an up to date list of which distros are ostree based would be more work than it's worth. I'd say that *in general* systems with `rpm-ostree` are most likely ostree based. This option just makes it possible for those exceptions to the rule to opt out of that assumption.

PR #722, opened 2021-06-02
Support winget (fix #721)

### LINKED ISSUE #721: Support for winget
[Winget](https://github.com/microsoft/winget-cli), a CLI for microsoft's AppInstaller which fills about the same role as `apt`, `brew`, `choco` etc. etc.

Windows 10

Check if `winget` is in PATH.

Use `winget upgrade --all` to upgrade all applications. Note that this can open graphical prompts, such as elevation requests and download windows. `-h` May be benefitial to supress some of these dialogs, at the expense of choosing defaults for the installations. [Here](https://docs.microsoft.com/en-us/windows/package-manager/winget/upgrade)'s the documentation for `winget upgrade`.

r-darwish:
## Standards checklist:

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -117,6 +117,7 @@ pub enum Step {
     Vagrant,
     Vcpkg,
     Vim,
+    Winget,
     Wsl,
     Yadm,
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -147,6 +147,7 @@ fn run() -> Result<()> {
     {
         runner.execute(Step::Chocolatey, "Chocolatey", || windows::run_chocolatey(&ctx))?;
         runner.execute(Step::Scoop, "Scoop", || windows::run_scoop(config.cleanup(), run_type))?;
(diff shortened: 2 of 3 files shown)


r-darwish:
@MCOfficer please test this change

MCOfficer:
[I keep getting 500s when adding comments so... you'll receive this whenever you'll receive it. Whack]

The topgrade code works, however winget seems to have some... funky behavior and a very healthy list of issues regarding upgrades. Since it also upgrades apps that weren't originally installed via winget, I can't really recommend enabling this step by default at this time. It already broke 3 things for me in the past week alone: Discord, OBS, and finally, [itself](https://github.com/microsoft/winget-pkgs/issues/15710).

OfficialEsco:
This comes from the dynamic link in the manifest and wingetbot automatically updating it and taking the File Version from the installer.
image

WinGet is using the version shown in the Control Panel which is why it thinks there is a update.
image

PR #729, opened 2021-06-08
PIP: avoid the warning 'pip is being invoked by an old script wrapper'

AlexanderWillner:
## Standards checklist:

## Short explanation

Under certain circumstances, e.g. on macOS 12, invoking `pip3` produces the following warning: `WARNING: pip is being invoked by an old script wrapper. This will fail in a future version of pip. Please see https://github.com/pypa/pip/issues/5599 for advice on fixing the underlying issue. To avoid this problem you can invoke Python with '-m pip' instead of running pip directly.`


MCOfficer:
What was wrong with this?

PR #734, opened 2021-06-27
Adding pnpm support

SilentVoid13:
## Standards checklist:

Adds support for [pnpm](https://github.com/pnpm/pnpm) global update.

2 questions:
- Should I check like for `npm` if `pnpm root -g` inside the user's home directory ? Like `yarn`, the default global directory is located at `~/pnpm-global`, so I don't think this shoud be necessary.
- Should I add the `--silent` flag ? I prefer without it to see what's going on, but maybe it's better if silent.

Thanks for this super useful program!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -299,6 +299,7 @@ fn run() -> Result<()> {
     runner.execute(Step::Vim, "voom", || vim::run_voom(&base_dirs, run_type))?;
     runner.execute(Step::Node, "npm", || node::run_npm_upgrade(&ctx))?;
     runner.execute(Step::Node, "yarn", || node::yarn_global_update(run_type))?;
+    runner.execute(Step::Node, "pnpm", || node::pnpm_global_update(run_type))?;
     runner.execute(Step::Deno, "deno", || node::deno_upgrade(&ctx))?;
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -299,6 +299,7 @@ fn run() -> Result<()> {
     runner.execute(Step::Vim, "voom", || vim::run_voom(&base_dirs, run_type))?;
     runner.execute(Step::Node, "npm", || node::run_npm_upgrade(&ctx))?;
     runner.execute(Step::Node, "yarn", || node::yarn_global_update(run_type))?;
+    runner.execute(Step::Node, "pnpm", || node::pnpm_global_update(run_type))?;
```
MCOfficer, line 302: Does pnpm act on the same stuff as node and yarn (i.e. ./.node-modules and global modules)? In that case you'd probably want to run pnpm first, lest npm does all the work but slower.

SilentVoid13: Hey @MCOfficer, yes `pnpm` is another javascript package manager like `npm` and `yarn`. However it doesn't make much sense to run them in a certain order, since they're separate package managers. `pnpm` and `npm` global updates are not linked and independent from one another.

PR #739, opened 2021-06-30
Make firmware upgrade configurable

ermshiperete:
This change adds an option, `firmware.upgrade`, that if set to `false` only checks for and displays available firmware updates. If set to `true` (default) the user is offered to run the firmware upgrade.

## Standards checklist:

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -88,4 +88,4 @@
 
 [firmware]
 # Offer to update firmware; if false just check for and display available updates
-#upgrade = false
+#upgrade = true
--- a/config.example.toml
+++ b/config.example.toml
@@ -85,3 +85,7 @@
 [npm]
 # Use sudo if the NPM directory isn't owned by the current user
 #use_sudo = true
+
+[firmware]
+# Offer to update firmware; if false just check for and display available updates
+#upgrade = false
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -730,6 +738,15 @@ impl Config {
             .unwrap_or(false)
     }
 
+    #[cfg(target_os = "linux")]
+    pub fn firmware_upgrade(&self) -> bool {
+        self.config_file
+            .firmware
+            .as_ref()
+            .and_then(|firmware| firmware.upgrade)
+            .unwrap_or(true)
```
r-darwish, line 747: I think that the default should be false since it's a dangerous operation

config.example.toml
```diff
@@ -85,3 +85,7 @@
 [npm]
 # Use sudo if the NPM directory isn't owned by the current user
 #use_sudo = true
+
+[firmware]
+# Offer to update firmware; if false just check for and display available updates
+#upgrade = false
```
r-darwish, line 91: Please put a newline after this line

r-darwish (changes requested):

src/config.rs
ermshiperete: I agree. `true` matches the current behavior, but `false` makes more sense :smile:

config.example.toml
ermshiperete: Done

r-darwish (approved):

PR #749, opened 2021-08-04
Add Android/Termux support

laurentlbm:
## Standards checklist:

- [x] *Optional:* I have tested the code myself _(manually, not through code)_

This PR adds support for Android, more specifically [Termux](https://termux.com/), a very popular terminal emulator and Linux environment. It updates Termux packages (using `pkg` and `apt`) and other already supported UNIX tools (on my device, that meant `cargo`, `npm`, `nvim`, `yadm`, `git repositories`, `tmux` (`TPM`) and `pip3`). It also correctly started in a new `tmux` session, if that config is enabled.

Building is done with `cargo build --target aarch64-linux-android`.

I have one question: rebooting is not useful in this case and only works on rooted devices anyway, should the code be modified to remove this option on Android?

Let me know if there is anything that can or should be improved.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -185,6 +185,9 @@ fn run() -> Result<()> {
         freebsd::upgrade_packages(sudo.as_ref(), run_type)
     })?;
 
+    #[cfg(target_os = "android")]
+    runner.execute(Step::Pkg, "Termux Packages", || android::upgrade_packages(&ctx))?;
+
     let emacs = emacs::Emacs::new(&base_dirs);
     if config.use_predefined_git_repos() {
         if config.should_run(Step::Emacs) {
--- a/src/steps/os/android.rs
+++ b/src/steps/os/android.rs
@@ -1,0 +1,31 @@
+use crate::execution_context::ExecutionContext;
+use crate::terminal::print_separator;
+use crate::utils::require;
+use anyhow::Result;
+
+pub fn upgrade_packages(ctx: &ExecutionContext) -> Result<()> {
+    let pkg = require("pkg")?;
+
+    print_separator("Termux Packages");
+
+    let mut command = ctx.run_type().execute(&pkg);
+    command.arg("upgrade");
+    if ctx.config().yes() {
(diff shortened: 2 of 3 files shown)


r-darwish (changes requested):
Seems like alot of code duplication between this and the debian upgrade function. Can you make a change so the the linux module will run as is when running in Termux?

laurentlbm:
You're right that there is code duplication. Termux uses a modified version of `apt`.
There are some reasons why I didn't use the debian code though:
- `sudo` is not needed in Termux for package installation, and is actually blocked.
- Termux recommends using their own command `pkg` for upgrades, which is a wrapper for `apt` (full explanation here: https://wiki.termux.com/wiki/Package_Management). 
  `pkg upgrade` does the following:
  1. **added**: updates mirror selection: *"Performs some client side repository load-balancing by automatically switching mirrors on a regular basis. That is important to prevent us hitting quota limit on hosting."*
  2. **same as `upgrade_debian`**: calls `apt update`
  3. **same as `upgrade_debian`**: calls `apt full-upgrade` (as opposed to `apt-get dist-upgrade`, but both are the same)
- I didn't want to apply the `apt_arguments` config since it is meant for Linux, and if a user shares their config between their computer and their phone, they might not want this config to automatically be applied.
  I felt that we could add a separate config in a future PR if the feature is requested.

The cleaning part of the debian upgrade is exactly the same though, and could be extracted and re-used by both.

About integrating Termux the Linux upgrade path for Termux, I don't think it would be worth it because most of it wouldn't apply:
1. again the `sudo` thing
2. OS detection can't be done like other Linux distributions (`/etc/os-release` doesn't exist)
3. `etc-update` doesn't exist
4. Brew doesn't support Termux
5. Rust considers Android a different OS, hence the build target (https://doc.rust-lang.org/reference/conditional-compilation.html#target_os)

You can read more about differences with Linux here: https://wiki.termux.com/wiki/Differences_from_Linux


So, what do you think I should do?

PR #750, opened 2021-08-05
Skip deno upgrade unless installed under .deno

tranzystorek-io:
## Standards checklist:

Deno can be installed from cargo or arch linux's repositories, so upgrade should only be called for user-wide shell-script installations.


src/main.rs
```diff
@@ -302,7 +302,7 @@ fn run() -> Result<()> {
     runner.execute(Step::Vim, "voom", || vim::run_voom(&base_dirs, run_type))?;
     runner.execute(Step::Node, "npm", || node::run_npm_upgrade(&ctx))?;
     runner.execute(Step::Pnpm, "pnpm", || node::pnpm_global_update(run_type))?;
-    runner.execute(Step::Deno, "deno", || node::deno_upgrade(&ctx))?;
+    runner.execute(Step::Deno, "deno", || node::deno_upgrade(&base_dirs, &ctx))?;
```
r-darwish, line 305: `base_dirs` should be contained in `ctx`

r-darwish (changes requested):

PR #768, opened 2021-10-14
Upgrade gnome shell extensions (fix  #755)

### LINKED ISSUE #755: Update gnome-shell extensions

It would be nice to update gnome-shell extensions using topgrade as well.

Gnome-shell

Linux

Check whether the DE is Gnome.

```
gdbus call --session --dest org.gnome.Shell.Extensions \ 
     --object-path /org/gnome/Shell/Extensions \
     --method org.gnome.Shell.Extensions.CheckForUpdates
```

**More information**

The actual upgrade will happen on the next login. This is the exact same behavior when updating via the gnome-extensions app.

r-darwish:
## Standards checklist:

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -84,6 +84,7 @@ pub enum Step {
     Gcloud,
     Gem,
     GitRepos,
+    GnomeShellExtensions,
     HomeManager,
     Jetpack,
     Krew,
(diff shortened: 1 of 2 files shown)


bdaase:
Alright, thanks!
Unfortunately I am not near my machine for a weeks right now, so I'm not sure when I will be able to test it...

PR #823, opened 2022-01-04
Re-add the Go step via go-global-update

qcloutier:
go-global-update (https://github.com/Gelio/go-global-update) is a small
tool to update all executables installed in a user's GOBIN, effectively
providing a `go get -u all` replacement for Go 1.16+.

Since it appears that Go will not be receiving a built-in way to do this
in the near future, this could be a good option in the meantime.

## Standards checklist:

### DIFF
--- a/src/steps/generic.rs
+++ b/src/steps/generic.rs
@@ -60,6 +60,27 @@ pub fn run_flutter_upgrade(run_type: RunType) -> Result<()> {
     run_type.execute(&flutter).arg("upgrade").check_run()
 }
 
+pub fn run_go(run_type: RunType) -> Result<()> {
+    let go = utils::require("go")?;
+    let gopath = run_type.execute(&go).args(&["env", "GOPATH"]).check_output()?;
+
+    print_separator("Go");
+
+    let go_global_update = utils::require("go-global-update")
+        .ok()
+        .or_else(|| PathBuf::from(gopath).join("bin/go-global-update").if_exists());
+    let go_global_update = match go_global_update {
+        Some(e) => e,
+        None => {
+            let message = String::from("go-global-update isn't installed so Topgrade can't upgrade Go packages.\nInstall go-global-update by running `go install github.com/Gelio/go-global-update@latest`");
+            print_warning(&message);
+            return Err(SkipStep(message).into());
+        }
+    };
+
+    run_type.execute(&go_global_update).check_run()
+}
+
 pub fn run_gem(base_dirs: &BaseDirs, run_type: RunType) -> Result<()> {
     let gem = utils::require("gem")?;
     base_dirs.home_dir().join(".gem").require()?;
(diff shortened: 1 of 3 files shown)


src/steps/generic.rs
```diff
@@ -60,6 +60,27 @@ pub fn run_flutter_upgrade(run_type: RunType) -> Result<()> {
     run_type.execute(&flutter).arg("upgrade").check_run()
 }
 
+pub fn run_go(run_type: RunType) -> Result<()> {
+    let go = utils::require("go")?;
+    let gopath = run_type.execute(&go).args(&["env", "GOPATH"]).check_output()?;
```
MCOfficer, line 65: Could use `std::env::var` here, unless i'm missing something.

qcloutier: This is necessary since users can define their `GOPATH` in their `GOENV` file instead of exporting it in their shell (see `go help environment`).

r-darwish:
I was looking into reimplementing the Go step myself and tried to figure out what's the correct solution for Go 1.16. I'm happy to see that someone made a tool for that, but at the moment I prefer only adding tools that gain some popularity. This specific project has 0 stars, so I'm looking for a tool which is considered more mainstream.

qcloutier:
Yeah, that's understandable. What are your thoughts on implementing the process in Topgrade itself? Effectively, it's just a matter of running `go version -m` (to get the source url) and `go install <url>@latest` on each item in the user's `GOBIN`. I think it's  foolproof, but there could be some gotchas.

jasikpark:
I am interested in this being solved! It seems like [`go-global-update`](https://github.com/Gelio/go-global-update/blob/a6ed616686f3d49c1ba5d0c9119974d30bd8d5d7/main.go) would be pretty simple to reproduce in rust

https://github.com/Gelio/go-global-update/blob/a6ed616686f3d49c1ba5d0c9119974d30bd8d5d7/main.go#L1-L98

r-darwish:
Still debating on this. It's possible to implement this logic in Topgrade itself but it feels like reinventing the wheel. On the other hand, `go-global-update` doesn't seem popular at all. I still don't understand how people in the Go community don't care enough about this issue to make sure a semi-official solution for this would exist.

r-darwish (changes requested):
I've decided to merge this instead of reinventing the wheel in Topgrade itself. If a more popular tool comes up we'll replace it. My only request is that instead of printing a warning we should silently skip this step if `go-gloval-update` doesn't exist. I'm not sure every go user would want to install this tool and I don't want to spam those who don't.

jasikpark:
I suppose users will either find the go-global-update on their own or find this PR when looking into how Topgrade supports Go? Or is there a list somewhere in the docs that describes which package managers are currently supported?

Gelio:
I've just found this PR and I'm super humbled by my project having an opportunity to be used in a well-known tool! Thank you for the opportunity

@jasikpark Looking at the wiki, there's [this page](https://github.com/r-darwish/topgrade/wiki/Step-list) that seems to be listing all the steps topgrade goes through, so I suppose go-global-update could be added there after this PR lands

PR #830, opened 2022-01-11
Enable updating Toolbx through topgrade

Funky185540:
## Standards checklist:

## Description

This adds a module to upgrade all toolboxes found on a host system.
[Toolbox][1] (More recently: `toolbx`) is a tool that easily provides
mutable, containerized command line environments. It is a means to work
and develop in a mutable environment on immutable operating systems such
as Fedora Silverblue, CoreOS, Kinoite and much more. However, it is not
limited to these use cases.

The integration into topgrade works as follows: Toolbx makes many of the
hosts folders available form inside the toolbx container by default. In
particular, all of the hosts filesystem is mapped under `/run/host`. This
module exploits this fact by:

- Gathering a list of toolboxes available on the host that runs
  `topgrade`
- Executing `toolbox run` for each of these toolboxes to run a command
  inside the toolboxes, **using the executing `topgrade` executable with
  the `--only system` arguments**

## Additional Notes

In my understanding of rust binaries, they are static, standalone executables. This is the one general assumption upon which this whole PR is based. Toolboxes are standalone containers, as such they don't run the same OS (version) as the host and the application probably cannot rely on having certain library versions available.

I do not know if this poses a problem here. Any feedback is highly appreciated!

[1]: https://containertoolbx.org/

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -276,6 +276,7 @@ fn run() -> Result<()> {
         runner.execute(Step::Sdkman, "SDKMAN!", || {
             unix::run_sdkman(&base_dirs, config.cleanup(), run_type)
         })?;
+        runner.execute(Step::Toolbx, "toolbx", || toolbx::run_toolbx(&ctx))?;
     }
 
     #[cfg(not(any(
--- a/src/main.rs
+++ b/src/main.rs
@@ -276,7 +276,6 @@ fn run() -> Result<()> {
         runner.execute(Step::Sdkman, "SDKMAN!", || {
             unix::run_sdkman(&base_dirs, config.cleanup(), run_type)
         })?;
-        runner.execute(Step::Toolbx, "toolbx", || toolbx::run_toolbx(&ctx))?;
     }
 
     #[cfg(not(any(
@@ -330,6 +329,7 @@ fn run() -> Result<()> {
 
     #[cfg(target_os = "linux")]
     {
+        runner.execute(Step::Toolbx, "toolbx", || toolbx::run_toolbx(&ctx))?;
         runner.execute(Step::Flatpak, "Flatpak", || linux::flatpak_update(&ctx))?;
         runner.execute(Step::Snap, "snap", || linux::run_snap(sudo.as_ref(), run_type))?;
         runner.execute(Step::Pacstall, "pacstall", || linux::run_pacstall(&ctx))?;
--- a/src/steps/mod.rs
+++ b/src/steps/mod.rs
@@ -7,6 +7,8 @@ pub mod powershell;
 pub mod remote;
 #[cfg(unix)]
 pub mod tmux;
+#[cfg(target_os = "linux")]
+pub mod toolbx;
 pub mod vim;
 #[cfg(unix)]
 pub mod zsh;
--- a/src/steps/toolbx.rs
+++ b/src/steps/toolbx.rs
@@ -1,0 +1,61 @@
+use anyhow::Result;
+
+use crate::config::Step;
+use crate::terminal::print_separator;
+use crate::{execution_context::ExecutionContext, utils::require};
+use log::debug;
+use std::path::Path;
+use std::{path::PathBuf, process::Command};
+
+fn list_toolboxes(toolbx: &Path) -> Result<Vec<String>> {
+    let output = Command::new(toolbx).args(&["list", "--containers"]).output()?;
+    let output_str = String::from_utf8(output.stdout)?;
+
+    let proc: Vec<String> = output_str
+        .lines()
+        // Skip the first line since that contains only status information
+        .skip(1)
+        .map(|line| match line.split_whitespace().nth(1) {
+            Some(word) => word.to_string(),
+            None => String::from(""),
+        })
+        .filter(|x| !x.is_empty())
+        .collect();
+
+    Ok(proc)
+}
+
+pub fn run_toolbx(ctx: &ExecutionContext) -> Result<()> {
+    let toolbx = require("toolbox")?;
+
+    print_separator("Toolbx");
+    let toolboxes = list_toolboxes(&toolbx)?;
+    debug!("Toolboxes to inspect: {:?}", toolboxes);
+
+    let mut topgrade_path = PathBuf::from("/run/host");
+    // Path of the running topgrade executable
+    // Skip 1 to eliminate the path root, otherwise push overwrites the path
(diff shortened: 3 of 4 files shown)


Funky185540:
Alright, thanks for clarifying!

I don't want to push anyone, but if you say that this has a *remote* chance of being accepted I would ask some other toolbox users to go test the feature and report if they find any issues with it. :)

MCOfficer:
> I don't want to push anyone, but if you say that this has a _remote_ chance of being accepted I would ask some other toolbox users to go test the feature and report if they find any issues with it. :)

Just to clarify: that decision belongs to r-darwish .

src/steps/toolbx.rs
```diff
@@ -0,0 +1,74 @@
...
+            None => String::from(""),
+        })
+        .filter(|x| !x.is_empty())
+        .collect();
+
+    Ok(proc)
+}
+
+pub fn run_toolbx(ctx: &ExecutionContext) -> Result<()> {
+    print_separator("Toolbx");
```
r-darwish, line 33: Put the separator after the `require`

src/steps/mod.rs
```diff
@@ -7,6 +7,8 @@ pub mod powershell;
 pub mod remote;
 #[cfg(unix)]
 pub mod tmux;
+#[cfg(unix)]
```
r-darwish, line 10: Isn't toolbox Linux specific?

src/steps/toolbx.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    let toolbx = require("toolbox")?;
+    let toolboxes = list_toolboxes(&toolbx)?;
+    debug!("Toolboxes to inspect: {:?}", toolboxes);
+
+    // Path of the running topgrade executable
+    let topgrade_path = String::from("/run/host") + std::env::current_exe()?.as_path().to_str().unwrap();
```
r-darwish, line 40: I think it's best to make a `PathBuf` from `/run/host` instead of turning everything to a string.

src/steps/toolbx.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    for tb in toolboxes.iter() {
+        print!("> Upgrading toolbox '{}'...", tb);
```
r-darwish, line 43: Instead of capturing the output of the inner topgrade you can set the environment variable `TOPGRADE_PREFIX` to the name of the container and just run it as usual, Topgrade will print the prefix in each step.

src/steps/toolbx.rs
```diff
@@ -0,0 +1,74 @@
...
+        // Write out immediately
+        io::stdout().flush()?;
+
+        let output = ctx
+            .run_type()
+            .execute(&toolbx)
+            .args(&["run", "-c", tb, &topgrade_path, "--yes", "--only", "system"])
```
r-darwish, line 50: Maybe pass `--yes` only if it was set in the current configuration?

src/steps/toolbx.rs
```diff
@@ -0,0 +1,74 @@
...
+
+        if let ExecutorOutput::Wet(output) = output {
+            let status = output.status;
+
+            if !status.success() {
+                println!(" FAILED");
+                // Is this ok?
+                io::stderr().write(&output.stderr).ok();
+
+                return Err(TopgradeError::ProcessFailed(status).into());
```
r-darwish, line 61: This will cancel remaining toolboxes, which is probably unwanted. I suggest that instead of returning you will initialize a boolean variable telling whether we failed or succeeded, and set it whenever a failure in one container occur. At the end of the function an error will be returned if any container failed.

r-darwish (changes requested):
Nice work. Please review my comments.

Funky185540:
Thanks for the feedback! Before I incorporate this I'm afraid I found an issue I'm not sure how to work around. It seems that the application is static except for the C runtime, `glibc` in my case. Unfortunately, toolbox containers running Fedora versions different from the host don't have a matching `libc` and can't find some symbols, so running topgrade there fails.

I tried linking the C runtime statically by adding this to `Cargo.toml`:

```
rustflags = ["-C", "target-feature=+crt-static"]
```

but it didn't solve the issue... Do you have an idea if there's a workaround to this?

r-darwish:
You should probably compile against MUSL instead of glibc.

src/steps/toolbx.rs
Funky185540: Fair enough, but then how exactly would I go about passing it to `Command::args`? To my knowledge it expects a slice of type `str`, which `PathBuf` doesn't satisfy.

Funky185540:
> You should probably compile against MUSL instead of glibc.

That works indeed, thanks! What exactly does this mean with regard to installing and running topgrade for Linux users?

src/steps/toolbx.rs
Funky185540: I'm afraid I'll need some help with that. Initializing and setting the boolean variable is straightforward, but:

- I don't know what to do then. Do I return some specially crafted `Err`, or how is this done "manually"? Because so far I could only gather from the code it's done implicilty by directly returning what falls out of e.g. `check_run`
- There's a critical error which I currently can't catch: If topgrade fails to run in a toolbox due to a glibc version mismatch, `toolbox run` will not return an error code but 0 instead. This of course gives the impression everything went well, which it didn't. Is there a way to get the stout/stderr of the executor outputs and parse those for the error message?

src/main.rs
```diff
@@ -276,6 +276,7 @@ fn run() -> Result<()> {
         runner.execute(Step::Sdkman, "SDKMAN!", || {
             unix::run_sdkman(&base_dirs, config.cleanup(), run_type)
         })?;
+        runner.execute(Step::Toolbx, "toolbx", || toolbx::run_toolbx(&ctx))?;
```
r-darwish, line 279: This should be moved to a Linux only section

r-darwish (changes requested):

r-darwish:
> > You should probably compile against MUSL instead of glibc.
> 
> That works indeed, thanks! What exactly does this mean with regard to installing and running topgrade for Linux users?

If you take a look at the release page you'll see that each release contains two Linux binaries for amd64, one compiling against glibc and another one, which is a completely static one, compiled against MUSL. You should download the MUSL version if you use a distribution such as Alpine Linux.

Funky185540:
Thanks for accepting it. :)

I'd like to note the information about using this with older Toolbx container versions somewhere, so others looking for it can find it and don't start opening issues when they run into it. Do you have a suggestion where to add this?

Issue #214, opened 2019-09-03
feature request: pull new docker images


RatanRSur:
Love the idea! I too have a script that intends to do this all for me so it's definitely a itch that needs scratching.

One thing my script does is pull new docker containers:
```
# do docker updates all in separate windows
for image in `docker images | tail -n +2 | cut -d ' ' -f 1 | sort | uniq | rg -v "<none>"`; do
    tmux split-window -d "docker pull $image; $update_layout"
    $update_layout
done
```
with one tmux pane for each update.

Perhaps topgrade could have an option that could pull the new containers? If you'd like me to open a PR, let me know and I'd be happy to.

r-darwish:
Pulling every docker image in the system is probably undesired for many people. If you wish to pull some specific repositories on your system you can add a custom command to the configuration.

PR #850, opened 2022-02-03
Upgrade podman/docker containers

Funky185540:
This PR implements automatic updates for all available ~~podman~~podman/docker containers in the user installation (not roots containers). 
I'm aware of the discussion in #214, so I made the code respect the tags that the containers were initially pulled with to ensure that no semver-breaking changes are performed when a specific tag was checked out initially (Like rust does with Cargo.toml).

The code is still very rough in a lot of places, but I wanted to put it up for discussion before finishing it entirely. If your opinion from #214 still holds, feel free to close the issue immediately.

## Standards checklist:

### DIFF
--- a/src/steps/containers.rs
+++ b/src/steps/containers.rs
@@ -1,15 +1,27 @@
 use anyhow::Result;
 
-use crate::error;
+use crate::error::{self, TopgradeError};
+use crate::executor::CommandExt;
 use crate::terminal::print_separator;
 use crate::{execution_context::ExecutionContext, utils::require};
-use log::debug;
+use log::{debug, error, warn};
 use std::path::Path;
 use std::process::Command;
+
+// A string found in the output of docker for containers that weren't found in
+// the docker registry. We use this to gracefully handle and skip containers
+// that cannot be pulled, likely because they don't exist in the registry in
+// the first place. This happens e.g. when the user tags an image locally
+// themselves or when using docker-compose.
+const NONEXISTENT_REPO: &str = "repository does not exist";
 
 /// Returns a Vector of all containers, with Strings in the format
 /// "REGISTRY/[PATH/]CONTAINER_NAME:TAG"
 fn list_containers(crt: &Path) -> Result<Vec<String>> {
+    debug!(
+        "Querying '{} images --format \"{{{{.Repository}}}}:{{{{.Tag}}}}\"' for containers",
+        crt.display()
+    );
     let output = Command::new(crt)
         .args(&["images", "--format", "{{.Repository}}:{{.Tag}}"])
         .output()?;
@@ -23,6 +35,7 @@ fn list_containers(crt: &Path) -> Result<Vec<String>> {
             continue;
         }
 
+        debug!("Using container '{}'", line);
         retval.push(String::from(line));
     }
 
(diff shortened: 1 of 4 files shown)


src/steps/podman.rs
```diff
@@ -0,0 +1,64 @@
...
+    for (registry, name, tag) in containers.iter() {
+        // FIXME: Should I return the whole string from podmans output directly instead?
+        let mut container = registry.to_owned();
+        container.push('/');
+        container.push_str(name);
+        container.push(':');
+        container.push_str(tag);
+        let args = vec!["pull", &container[..]];
+
+        // TODO: How would I check whether the execution was successful?
```
r-darwish, line 52: `check_run()` does it. However, it will cause the function to quit and skip the remaining contains if one pull fails. I suggest that you keep a boolean variable indicating whether any container failed and try to iterate all of them even if some fail.

src/steps/podman.rs
```diff
@@ -0,0 +1,64 @@
...
+        ctx.run_type().execute(&podman).args(&args).check_run()?;
+    }
+
+    // Remove dangling images
+    // TODO: Should prune be automatic, or should it respect the "--yes" flag?
```
r-darwish, line 57: There's a specific flag for running cleanups

src/steps/podman.rs
```diff
@@ -0,0 +1,64 @@
...
+
+pub fn run_podman(ctx: &ExecutionContext) -> Result<()> {
+    let podman = require("podman")?;
+
+    print_separator("Podman");
+    let containers = list_containers(&podman)?;
+    debug!("Containers to inspect: {:?}", containers);
+
+    for (registry, name, tag) in containers.iter() {
+        // FIXME: Should I return the whole string from podmans output directly instead?
```
r-darwish, line 44: If all you do is join it back in then I suggest yes.

src/steps/podman.rs
```diff
@@ -0,0 +1,64 @@
...
+
+/// Returns a Vector of all containers, as a triple (repository, container name, version)
+fn list_containers(podman: &Path) -> Result<Vec<(String, String, String)>> {
+    let output = Command::new(podman).args(&["images", "--format", "json"]).output()?;
+    let output_str = String::from_utf8(output.stdout)?;
+    let json_str = json::parse(&output_str)?;
+
+    let mut retval = vec![];
+    for lentry in json_str.members() {
+        // FIXME: Bad unwraps!
```
r-darwish, line 18: Maybe parsing the JSON with serde will be better.

src/steps/podman.rs
```diff
@@ -0,0 +1,64 @@
...
+use std::process::Command;
+
+/// Returns a Vector of all containers, as a triple (repository, container name, version)
+fn list_containers(podman: &Path) -> Result<Vec<(String, String, String)>> {
+    let output = Command::new(podman).args(&["images", "--format", "json"]).output()?;
+    let output_str = String::from_utf8(output.stdout)?;
+    let json_str = json::parse(&output_str)?;
+
+    let mut retval = vec![];
+    for lentry in json_str.members() {
```
r-darwish, line 17: I suggest  you debug print every line.

r-darwish (changes requested):

r-darwish:
@Funky185540 docker support has been requested in #853. Can you make this PR support both docker and podman?

Funky185540:
> @Funky185540 docker support has been requested in #853. Can you make this PR support both docker and podman?

Can't make any promises, but that was the idea for the long run anyways. I'll look into it!

src/steps/podman.rs
Funky185540: Done! I now make it return a `StepFailed` at the end if any of the partial steps failed. It now also prints debug outputs if something goes wrong. Should I maybe use `warning!` or `error!` for things that cause step failures? Do you have any rules for this?

src/steps/podman.rs
Funky185540: Much the same API in fact, but as `serde` is already a dependency, I now changed it to use `serde_json` rather than `json` (Which still adds a new dependency, unfortunately). I'll have to replace the JSON parsing with something else in any case to make it compatible with docker. I'll try my luck on this later today.

Funky185540:
Theoretically it should now work with both podman and docker. Once I know what to do about the use of `debug!/warning!/error!` and I have the feedback from #853 this is good to go from my point of view. :)

src/steps/containers.rs
```diff
@@ -0,0 +1,113 @@
...
+                    Some(TopgradeError::ProcessFailedWithOutput(_, stderr)) => {
+                        if stderr.contains(NONEXISTENT_REPO) {
+                            true
+                        } else {
+                            return Err(e);
+                        }
+                    }
+                    _ => return Err(e),
+                },
+            } {
```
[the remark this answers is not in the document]
r-darwish: I think it can be made much simpler if no error will cause the function to quit. This is similar to what we do when pulling git repositories. If any repo fails updating then `success` will be made `false` and the step will eventually be marked as failed, but I don't think we should stop trying the rest of the repos in any case.

src/steps/containers.rs
```diff
@@ -0,0 +1,113 @@
...
+        debug!("Using container '{}'", line);
+        retval.push(String::from(line));
+    }
+
+    Ok(retval)
+}
+
+pub fn run_containers(ctx: &ExecutionContext) -> Result<()> {
+    // Prefer podman, fall back to docker if not present
+    let crt = match require("podman") {
```
r-darwish, line 53: You can do this with `.or_else`

r-darwish (changes requested):

src/steps/containers.rs
Funky185540: Correct!

Funky185540:
> I think it can be made much simpler if no error will cause the function to quit. This is similar to what we do when pulling git repositories. If any repo fails updating then success will be made false and the step will eventually be marked as failed, but I don't think we should stop trying the rest of the repos in any case.

That's absolutely right, and it was the initial idea in any case, hence I introduced the `success` variable. I just ignored that while working out the error handling... Additional benefit: This makes the whole error handling routine much more compact

PR #858, opened 2022-02-12
Support multiple WSL distributions (fix #855)

### LINKED ISSUE #855: Option (or default behavior) to update all WSL distros, not just default
I was amazed to find out that topgrade updates WSL on Windows as long as the distro has it installed but it only does that for the one that is set as default since it runs the `wsl` command without additional distro parameters.

From what I've seen in the source code the command
`wsl bash -c exec topgrade`
or something along those lines is executed after checking wsl and probing for topgrade.

One could imagine running a
`wsl --list`
first and piping this into a loop iterating over all found distros (~foreach Distro in pseudo-code) executing the step commands with a
`wsl -d <DISTRONAME>`
prefix.

Don't know if this would be the desired behavior for everyone but as someone who has quite a few distros installed on WSL it would be quite useful to me at least as an option.

r-darwish:


r-darwish:
@Lim-Dul can you test this feature?

Lim-Dul:
Yes I could. Do I have to compile from source or are there automated development builds available somewhere?

r-darwish:
Just clone this branch and run `cargo run`

r-darwish:
I tried switching it to `--` but it didn't detect Topgrade. I think that bash don't reload `.profile` when `--` is used. If this works as expected then I'll merge it as it is

PR #874, opened 2022-03-16
steps: vim: Show outputs, simplify error handling

har7an:
Print stdout and stderr directly to the invoking terminal, rather than
collecting them and only printing them when the update failed.
Simplifies error handling by having the executor check for the command
exit status and automatically handling dry runs.

*Background*: I use `dein` as plugin manager for vim and had setup a custom update command for that in my topgrade config. I forgot to remove it after #857 was merged, so it stuck around. Recently I discovered that `vim-airline` [had a rewrite in its git-history](https://github.com/vim-airline/vim-airline/issues/2515), causing the update to fail but without `dein` returning an error exit status (i.e. it still returned 0). Thus the `vim` upgrade step would report "Plugins updated", whereas in the custom commands output I could see that it failed to upgrade one plugin. Since the other steps (at least the ones I use) also don't work silently (i.e. without showing me stdout/stderr), I decided to update the `vim` step to behave the same.

## Standards checklist:

### DIFF
--- a/src/steps/vim.rs
+++ b/src/steps/vim.rs
@@ -1,7 +1,7 @@
-use crate::error::{SkipStep, TopgradeError};
+use crate::error::SkipStep;
 use anyhow::Result;
 
-use crate::executor::{CommandExt, ExecutorOutput, RunType};
+use crate::executor::{CommandExt, RunType};
 use crate::terminal::print_separator;
 use crate::{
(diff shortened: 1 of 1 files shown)


r-darwish:
As I recall this change was made because most plugin managers would output garbage. Is there any way to change the vim script so that it properly returns 1 when dein fails?

PR #876, opened 2022-03-23
Use `docker image ls` instead of `docker images` to list images

tuxiqae:
Use the same format for all `docker image` commands

## Standards checklist:


r-darwish:
@tuxiqae does this work for Podman as well?

woohgit:
> @tuxiqae does this work for Podman as well?

Yes it does.

tuxiqae:
> @tuxiqae does this work for Podman as well?

Yup, not only does the documentation mention this, I also tested it myself and it seems to be working perfectly well.

PR #879, opened 2022-03-25
Fix compilation on Debian Buster

beckerj:
Setting tokio to 1.16.0 which is the last version compatible with debian buster rustc:
tokio-rs/tokio#4491

## Standards checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1758,7 +1758,7 @@ checksum = "cda74da7e1a664f795bb1f8a87ec406fb89a02522cf6e50620d016add6dbbf5c"
 
 [[package]]
 name = "tokio"
-version = "1.17.0"
+version = "1.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2af73ac49756f3f7c01172e34a23e5d0216f6c32333757c2c61feb2bbff5a5ee"
 dependencies = [


r-darwish:
Sorry, but if tokio choose to stop supporting Buster then there's no reason for Topgrade to downgrade a dependency. Topgraded is supported in the latest versions of  Debian and the latest Ubuntu LTS.

beckerj:
Sorry, that was actually a mistake on my part that i now corrected. The issue is with debian bullseye, the current stable version:
Debian 11 (bullseye) comes with rustc Version: 1.48.0+dfsg1-2
Changelog of tokio 1.17.0: update minimum supported Rust version to 1.49 (https://github.com/tokio-rs/tokio/pull/4457)

MCOfficer:
I'd argue that this is ok - end-users can always use the precompiled binary, and developers should use rustup anyways.

beckerj:
The point here is, every install of topgrade that was installed using cargo install on debian bullseye fails right now, during the update of topgrade.
If you say I now need to upgrade it manually by downloading the binary that of course works for me, but somebody else might not know what the error means.
Also, I thought not having to do manual updates was the whole reason for topgrades existence.

Is there any reason tokio should not be 1.46?

MCOfficer:
Having users get an unknown error is unfortunate for sure. We might be able to detect our MRSV and throw a warning or panic using [rustc-version](https://crates.io/crates/rustc_version) in the buildscript, that might be a workaround.

As for manually upgrading - no worries, topgrade will upgrade itself, so that should be ok.

> Is there any reason tokio should not be 1.16?

None I can see immediately, but it would set a precedent about our MRSV. So far I believe we've stuck to whatever's on the oldest Ubuntu LTS. We (read: Roey) can now make the call whether to bind ourselves to debian as well, or drop it because ugh.

beckerj:
> As for manually upgrading - no worries, topgrade will upgrade itself, so that should be ok.

Unfortunately, this is not the case. The cargo install version runs with:
`2022-03-28T07:46:41.830Z DEBUG topgrade         > Self Update: false`
which is **not** configured in topgrade.toml, so I assume it is default

MCOfficer:
> > As for manually upgrading - no worries, topgrade will upgrade itself, so that should be ok.
> 
> Unfortunately, this is not the case. The cargo install version runs with: `2022-03-28T07:46:41.830Z DEBUG topgrade > Self Update: false` which is **not** configured in topgrade.toml, so I assume it is default

The cargo version updates itself via cargo, the binary version should update itself. I use the binary for the most part, and it works fine.

beckerj:
Problem just being that any installation of topgrade via cargo on debian is practically broken right now. 
I get that you might not want to support cargo on debian, but maybe then there should be at least a note that installation should only be done by using the binary on debian.

MCOfficer:
> Problem just being that any installation of topgrade via cargo on debian is practically broken right now. I get that you might not want to support cargo on debian, but maybe then there should be at least a note that installation should only be done by using the binary on debian.

Agreed, there should be some way of knowing beyond the slightly hidden mention of "Rust 1.51" in the README. I'll look into perhaps spewing out a warning/error at compile-time.

PR #914, opened 2022-05-01
Replace the apt step with nala if installed.

### LINKED ISSUE #897: Replace the apt step with nala if installed.
It's about "[nala](https://salsa.debian.org/volian-team/nala)" an `apt` frontend replacement with optimized terminal output and added features like automatically choosing the best mirror or parallelized downloads for faster updates.
Any Linux distro that uses apt.  
### What should Topgrade do to figure out if the tool needs to be invoked?
I think checking for the `nala` command in the PATH should be sufficient. 
Running `sudo nala upgrade` does the job.

r-darwish:
Fixes #897

## Standards checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,10 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+      - master
 
 jobs:
   build:
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -32,7 +32,7 @@
         {
             "type": "promptString",
             "id": "step",
-            "description": "step nname",
+            "description": "step name",
         }
     ]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,7 +7,7 @@
         {
             "type": "lldb",
             "request": "launch",
-            "name": "Debug executable 'topgrade'",
+            "name": "Topgrade",
             "console": "integratedTerminal",
             "cargo": {
                 "args": [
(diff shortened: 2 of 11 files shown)


r-darwish:
@RayJW can you please test this branch?

RayJW:
At the moment, it updates the sources with `nala update` however it fails at: 

`Usage: nala [OPTIONS] COMMAND [ARGS]...
Try 'nala -h' for help.

Error: No such command 'dist-upgrade'.
`

This is because the commands don't map 1 to 1 with apt. It would be sufficient to just run `nala upgrade` as this updates the sources before running a system upgrade.

r-darwish:
@RayJW I pushed a fix. Can you please try again?

RayJW:
@r-darwish The upgrade process now works, however since upgrade also involves an update, the sources now get updated twice which is a waste of time. The easiest solution would be to run `sudo nala upgrade --no-update` so there don't have to be more checks to decide if running an update is necessary. Alternatively, `sudo nala update` can be skipped and only upgrade without any flags should be run. 
Either solution would work.

r-darwish:
> @r-darwish The upgrade process now works, however since upgrade also involves an update, the sources now get updated twice which is a waste of time. The easiest solution would be to run `sudo nala upgrade --no-update` so there don't have to be more checks to decide if running an update is necessary. Alternatively, `sudo nala update` can be skipped and only upgrade without any flags should be run. Either solution would work.

I pushed a fix to that as well

RayJW:
@r-darwish Everything works as expected now. 
Maybe it would actually make sense to use the second solution of skipping `sudo nala update` and then run `sudo nala upgrade` because now the message reporting how many updates are available is duplicated. However, this is more of an improvement and not critical. Thanks for your work either way!

RayJW:
@r-darwish Now everything works like a charm. Thank you for your work!

PR #920, opened 2022-05-06
Desktop notifications show step in summary

### LINKED ISSUE #918: Desktop notifications show step in summary
## What did you expect to happen?

Notification summary is "Topgrade".

The macos section of `notify_desktop` seems to imply this intent as well, listing "Topgrade" as the summary and the body as the message.

https://github.com/r-darwish/topgrade/blob/5d168bbf0d99040eed101cdc3a83f958781193d0/src/terminal.rs#L91-L99

## What actually happened?

Topgrade is not mentioned, and step name is given as summary.

This is clear from the linux section of `notify_desktop`, where "Topgrade" is given as the app name (imo it should be "topgrade", just like on macos) and the message is given as the first positional argument to `notify-send`.

https://github.com/r-darwish/topgrade/blob/5d168bbf0d99040eed101cdc3a83f958781193d0/src/terminal.rs#L101-L110

From `notify-send --help`

```
Usage:
  notify-send [OPTION…] <SUMMARY> [BODY] - create a notification
Application Options:
  -a, --app-name=APP_NAME           Specifies the app name for the icon
```

As is shown here, the summary is the first positional argument and the body is the second positional argument.

Also: Why aren't we using notify-rust for linux notifications? The library has linux support, right?

## Additional details
  Manjaro Linux XFCE
  Manjaro community repo (Topgrade 8.3.0)

r-darwish:
Fixes #918

## Standards checklist:

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -104,7 +104,7 @@ impl Terminal {
                         command.arg("-t");
                         command.arg(format!("{}", timeout.as_millis()));
                     }
-                    command.args(&["-a", "Topgrade"]);
+                    command.args(&["-a", "Topgrade", "Topgrade"]);
                     command.arg(message.as_ref());
                     command.output().ok();
                 }


r-darwish:
@Spaceface16518 please test this change

PR #925, opened 2022-05-07
Support deb-get

### LINKED ISSUE #923: Support deb-get
deb-get gives apt-get functionality for .debs published in 3rd party repositories or via direct download https://github.com/wimpysworld/deb-get
Linux Ubuntu and derivative distributions. 
### What should Topgrade do to figure out if the tool needs to be invoked?
Check if /bin/deb-get exists
sudo deb-get update
sudo deb-get upgrade

r-darwish:
Fixes #923

## Standards checklist:

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,14 @@
+{
+	"version": "2.0.0",
+	"tasks": [
+		{
+			"type": "cargo",
+			"command": "clippy",
+			"problemMatcher": [
+				"$rustc"
+			],
+			"group": "test",
+			"label": "rust: cargo clippy"
+		}
+	]
+}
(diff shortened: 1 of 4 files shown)


r-darwish:
@Zahrun please test this change

Zahrun:
Thank you for the quick addition.
It seems the "update" step is not required since [the only difference](https://github.com/wimpysworld/deb-get/blob/cda87d9123377853ebbf36809c87d20e239dc08d/deb-get#L316) is that it will do "apt-get update", and we already do this in a previous topgrade step . So "upgrade" alone will work for deb-get.
Also we are missing this header:
`―― 10:40:35 - deb-get ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――`

EDIT: you corrected the title before I posted. Let me test it.

Zahrun:
The separator displays correctly.
We can remove the `ctx.execute_elevated(&deb_get, false)?.arg("update").check_run()?;` line and we are good.

PR #940, opened 2022-05-23
Trim go paths

r-darwish:


r-darwish:
@jtsagata can you please check this branch?

MCOfficer:
Files and directories can technically end in whitespace, so there's the extremely unlikely scenario that this will break go updates for users who have their gopath set to, say `~/mygo `

jtsagata:
It's working as expected with the trimming.

Now, if my $GOPATH, or any directory,  ends with space, I will expect much more Sirius problems than this :-). Maybe just remove a trailing `\n;

PR #942, opened 2022-05-26
Stop using packages.el

r-darwish:


r-darwish:
@stefanhusmann can you test this branch?

stefanhusmann:
Works fine here, but maybe also a user of bare package.el shoud test it also.

PR #949, opened 2022-06-01
npm/cli 8.11.0 deprecated -g flags

markusgod:
As of https://github.com/npm/cli/releases/tag/v8.11.0 npm deprecated `-g` syntax replaced with `--location=global`

```
―― 16:27:28 - Node Package Manager ―――――――――――――――――――――――――――――――――――――――――――――
npm WARN config global `--global`, `--local` are deprecated. Use `--location=global` instead.
```
## Standards checklist:

### DIFF
--- a/src/steps/node.rs
+++ b/src/steps/node.rs
@@ -10,6 +10,7 @@ use directories::BaseDirs;
 use log::debug;
 #[cfg(unix)]
 use nix::unistd::Uid;
+use semver::Version;
 
 use crate::executor::{CommandExt, RunType};
 use crate::terminal::print_separator;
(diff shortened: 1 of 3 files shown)


src/steps/node.rs
```diff
@@ -32,23 +33,42 @@ impl NPM {
...
+        let version = self.version()?;
+        let args = if version < Version::new(8, 11, 0) {
+            ["update", "-g"]
+        } else {
+            ["update", "--location=global"]
+        };
         if use_sudo {
             run_type
                 .execute("sudo")
                 .arg(self.pnpm.as_ref().unwrap_or(&self.command))
```
r-darwish, line 67: Do you know if this will work with pnpm?

r-darwish (commented):
I didn't know about the semver module, thank you!
Could you just check or tell me if this will work if a new version of NPM is installed but also pnpm?

src/steps/node.rs
markusgod: I've added a check for pnpm in https://github.com/r-darwish/topgrade/pull/949/commits/c9aebd33ca506683f91bacf4a320fd8a553f7993, it should use old arguments.

r-darwish (approved):

r-darwish:
> Oh, I completely missed it. It works fine on my machine BUT, it looks like pnpm update is not running at all 😅. It shouldn't since pnpm does not yet have that keys. I'll update pr when I'll have more time. In the mean time why was pnpm moved out of its own step?

The reasoning behind it is that pnpm should be a replacement to npm, not run together with it.

PR #954, opened 2022-06-14
Ensure `selfupdate` is enabled for SDKMAN!

qcloutier:
This subcommand is unavailable when the `sdkman_selfupdate_feature`
option is disabled, as is the case when SDKMAN! is installed via
Homebrew.

https://github.com/sdkman/sdkman-cli/pull/1042

## Standards checklist:

### DIFF
--- a/src/steps/os/unix.rs
+++ b/src/steps/os/unix.rs
@@ -8,6 +8,7 @@ use crate::utils::{require, PathExt};
 use crate::Step;
 use anyhow::Result;
 use directories::BaseDirs;
+use ini::Ini;
 use log::debug;
 use std::fs;
 use std::os::unix::fs::MetadataExt;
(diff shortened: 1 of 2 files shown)


src/steps/os/unix.rs
```diff
@@ -343,11 +344,28 @@ pub fn run_sdkman(base_dirs: &BaseDirs, cleanup: bool, run_type: RunType) -> Res
...
-        .execute(&bash)
-        .args(&["-c", cmd_selfupdate.as_str()])
-        .check_run()?;
+    let sdkman_config_path = env::var("SDKMAN_DIR")
+        .map(PathBuf::from)
+        .unwrap_or_else(|_| base_dirs.home_dir().join(".sdkman"))
+        .join("etc")
+        .join("config")
+        .require()
+        .map(|p| format!("{}", &p.display()))?;
```
MCOfficer, line 353: You can probably remove this mapping since [`Ini::load_from_file`](https://docs.rs/rust-ini/latest/ini/struct.Ini.html#method.load_from_file) takes `AsRef<Path>`

src/steps/os/unix.rs
```diff
@@ -343,11 +344,28 @@ pub fn run_sdkman(base_dirs: &BaseDirs, cleanup: bool, run_type: RunType) -> Res
...
+
+    let sdkman_config = Ini::load_from_file(sdkman_config_path)?;
+    let selfupdate_enabled: String = sdkman_config
+        .general_section()
+        .get("sdkman_selfupdate_feature")
+        .unwrap_or("false")
+        .into();
```
MCOfficer, line 360: I believe this can stay a `&str` or even be cast as bool, but it is difficult to see borrow checker drama just by looking at the code.

MCOfficer (changes requested):
You also need to depend on `rust-ini` in `Cargo.toml`'s MacOS section.

PR #964, opened 2022-06-27
Fix doom emacs upgrading (fix #961)

### LINKED ISSUE #961: Use secondary check to see if Doom Emacs upgrade should be run
Topgrade should use a secondary check to determine if Doom Emacs upgrade should be run.  Given that a default installation of Doom Emacs does NOT produce an init.el file in the ~/.emacs.d directory, the check for upgrading Emacs fails (which means it never attempts to detect if Doom Emacs is present or not.  For Doom Emacs, this actually lives under ~/.doom.d/init.el.

vmalloc:


DottoDev:

PR #982, opened 2022-07-28
Add new step: GNU Guix

### LINKED ISSUE #979: Support GUIX: distro independent, declarative, package manager
## I want to suggest a new step
I would like support for the guix package manager, its my favorite declarative/source based package manager. 
And i use it on all my devices.

To learn what it's about, info [**here**](https://guix.gnu.org/en/about/)

For convenience, I have put a mirror of the git repo here  [**here**](https://github.com/guix-mirror/guix):
Reason why its the mirror is some people may find the GitHub interface more families (README file pops up, etc). But the original repo is listed there.

Any GNU/Linux System

#### Check if the package manager is installed.

This checks if guix can be used, that the tool is in $PATH:
`if [ $(which guix) ]; then echo 1; else echo 0; fi`
This will be the first check.

`guix pull` Upgrades guix package manager & checks for package updates of installed packages
[More Info...](https://guix.gnu.org/manual/en/html_node/Invoking-guix-pull.html)

`guix package -u` Upgrades packages installed by the user, which are located in their `.guix_profile`
[More Info...](https://guix.gnu.org/manual/en/html_node/Invoking-guix-package.html)

## I want to suggest some general feature
Topgrade could optionally also check if an attempted installation happened, but failed, but not essential.

This checks if guix is installed at all(even if installed incorrectly):
`if [ -e /gnu ]; then echo 1; else echo 0; fi`
The reason being, is that any time guix is installed, this directory will be made.
Potentially, it may be useful to tell the user that guix was installed incorrectly, but i don't know if this is in scope of this issue, so its just a side note

## More information

Yes, i would do a git clone of that branch, then cargo run, to test it i believe.

I do not know the rust programming language, but I can at least do some exploratory testing and share results.

JamesClarke7283:
Adds support for GNU Guix package manager.

Closes #979
Needs better error handling/messages. But it functions when ive tested it & if the step fails, the program continues as normal.
No custom error messages, happen yet, at the moment error output is just the default from the package manager. see the prior mentioned issue for more details.

I have run the checks, on check that failed, was `cargo clippy`. because of a unneeded return statement, i will fix that warning later. Sorry if my code is not great, i have no experience with rust, until now. Feedback on what to improve would be much appreciated. 

## Standards checklist:

## Features
- [x] Update guix (both the database, and the guix binary itself)
- [x] Upgrade guix packages


JamesClarke7283:
Cargo Clippy gives no warnings caused by my changes (checked upstream master and was consistent with that), and I removed the custom Error/Messages handling item, from the requirements.

Had help from Leo Spratt (@enchant94) who knows a lot more about rust specific best practices.

strangelittlemonkey:
Please remove the swp file that was accidentally added to the commit.

JamesClarke7283:
> Please remove the swp file that was accidentally added to the commit.

Thanks for letting me know, do let me know if anything needs to be improved, I had no experience with programming in Rust before this, so I made the decision to not add the extra messages.

Do you think it would be worthwhile to add a printout, saying what command to run to install the GNU Guix package manager?

If so, Would the method to install GNU Guix stated here, be too long?
https://guix.gnu.org/manual/en/html_node/Binary-Installation.html#example

JamesClarke7283:
Does anything else need doing for this to be ready for merge?

DottoDev:
Link: https://github.com/DottoDev/topgrade/pull/1

PR #996, opened 2022-08-23
feat: add support for bun

glc-mvion:
Hello ! :wave: 

I took the liberty to add support for [bun](https://github.com/oven-sh/bun), a recent and promising JavaScript tool and runtime.

It features self update, so I was able to derive existing work in the repository.

One thing I was not sure about is in which `cfg` part in `config.rs` I should add my step since Bun is only available for Linux and macOS at this time. If I put it in the wrong part, please indicate me, I will move i tto the right place.

## Standards checklist:

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -71,6 +71,7 @@ pub enum Step {
     Atom,
     BrewCask,
     BrewFormula,
+    Bun,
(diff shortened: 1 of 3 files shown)


DottoDev:

